PR #80, opened 2014-09-16
Create StatementSerializer and StatementDeserializer

Benestar:
This patch adopts the changes in Statement which uses composition instead of inheritance for claims. It depends on some future version of WikibaseDataModel which supports claims as parameters for the Statement constructor.

See also http://git.io/wXZTag


Benestar:
This is actually done in #115

PR #85, opened 2014-09-19
More clearly document package private

JeroenDeDauw:


thiemowmde:
Why that self-merge? This is almost the opposite of the proposal in #101. Pinging @brightbyte since we discussed this some time ago.

PR #97, opened 2014-11-04
Add requirements for language fall backs.

JanZerebecki:
Requires https://github.com/wmde/WikibaseDataModel/pull/272

### DIFF
--- a/src/Deserializers/EntityDeserializer.php
+++ b/src/Deserializers/EntityDeserializer.php
@@ -4,6 +4,7 @@ namespace Wikibase\DataModel\Deserializers;
 
 use Deserializers\Deserializer;
 use Deserializers\Exceptions\DeserializationException;
+use Deserializers\Exceptions\InvalidAttributeException;
 use Deserializers\TypedObjectDeserializer;
 use Wikibase\DataModel\Entity\Entity;
 use Wikibase\DataModel\Statement\StatementList;
@@ -99,8 +100,8 @@ abstract class EntityDeserializer extends TypedObjectDeserializer {
 	private function deserializeValuePerLanguageSerialization( array $serialization ) {
 		$array = array();
 
-		foreach( $serialization as $valueSerialization ) {
-			$this->assertIsValidValueSerialization( $valueSerialization );
+		foreach( $serialization as $requestedLanguage => $valueSerialization ) {
+			$this->assertIsValidValueSerialization( $valueSerialization, $requestedLanguage );
 			$array[$valueSerialization['language']] = $valueSerialization['value'];
 		}
 
@@ -113,24 +114,46 @@ abstract class EntityDeserializer extends TypedObjectDeserializer {
 		}
 		$this->assertAttributeIsArray( $serialization, 'aliases' );
 
-		foreach( $serialization['aliases'] as $aliasesPerLanguageSerialization ) {
+		foreach( $serialization['aliases'] as $requestedLanguage => $aliasesPerLanguageSerialization ) {
 			if ( !is_array( $aliasesPerLanguageSerialization ) ) {
 				throw new DeserializationException( "Aliases attribute should be an array of array" );
 			}
 
 			foreach( $aliasesPerLanguageSerialization as $aliasSerialization ) {
-				$this->assertIsValidValueSerialization( $aliasSerialization );
+				$this->assertIsValidValueSerialization( $aliasSerialization, $requestedLanguage );
 				$entity->addAliases( $aliasSerialization['language'], array( $aliasSerialization['value'] ) );
 			}
 		}
 	}
 
-	private function assertIsValidValueSerialization( $serialization ) {
+	private function assertNotAttribute( array $array, $key ) {
+		if ( array_key_exists( $key, $array ) ) {
+			throw new InvalidAttributeException(
+				$key,
+				$array[$key],
+				'Deserialization of attribute ' . $key . ' not supported.'
+			);
+		}
+	}
+
+	private function assertRequestedAndActualLanguageMatch( $serialization, $requestedLanguage ) {
+		if ( $serialization['language'] !== $requestedLanguage ) {
+			throw new DeserializationException(
+				'Deserialization of a value of the attribute language (actual)'
+					. ' that is not matching the language key (requested) is not supported: '
+					. $serialization['language'] . ' !== ' . $requestedLanguage
+			);
+		}
+	}
+
+	private function assertIsValidValueSerialization( $serialization, $requestedLanguage ) {
 		$this->requireAttribute( $serialization, 'language' );
 		$this->requireAttribute( $serialization, 'value' );
+		$this->assertNotAttribute( $serialization, 'source' );
 
 		$this->assertAttributeInternalType( $serialization, 'language', 'string' );
 		$this->assertAttributeInternalType( $serialization, 'value', 'string' );
+		$this->assertRequestedAndActualLanguageMatch( $serialization, $requestedLanguage );
 	}
 
 	private function setClaimsFromSerialization( array $serialization, Entity $entity ) {
--- a/src/Serializers/FingerprintSerializer.php
+++ b/src/Serializers/FingerprintSerializer.php
@@ -6,6 +6,12 @@ use Serializers\Exceptions\SerializationException;
 use Serializers\Exceptions\UnsupportedObjectException;
 use Serializers\Serializer;
 use Wikibase\DataModel\Entity\Entity;
+use Wikibase\DataModel\Term\AliasGroup;
+use Wikibase\DataModel\Term\AliasGroupFallback;
+use Wikibase\DataModel\Term\AliasGroupList;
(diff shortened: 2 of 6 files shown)


src/Serializers/FingerprintSerializer.php
```diff
@@ -47,25 +53,22 @@ public function addIdToSerialization( Entity $entity, array &$serialization ) {
 	}
 
 	public function addLabelsToSerialization( Entity $entity, array &$serialization ) {
-		$labels = $entity->getLabels();
+		$labels = $entity->getFingerprint()->getLabels();
```
thiemowmde: Why do you need "the actual object" and what does "actual" mean in the first place?

JanZerebecki: Actual means here the one that contains the necessary information. $entity->getLabels(); just creates a new array with a subset of the information which is not enough here.

src/Deserializers/EntityDeserializer.php
```diff
@@ -113,24 +114,46 @@ private function setAliasesFromSerialization( array $serialization, Entity $enti
...
+	}
+
+	private function assertIsValidValueSerialization( $serialization, $requestedLanguage ) {
 		$this->requireAttribute( $serialization, 'language' );
 		$this->requireAttribute( $serialization, 'value' );
+		$this->assertNotAttribute( $serialization, 'source' );
 
 		$this->assertAttributeInternalType( $serialization, 'language', 'string' );
 		$this->assertAttributeInternalType( $serialization, 'value', 'string' );
+		$this->assertRequestedAndActualLanguageMatch( $serialization, $requestedLanguage );
```
thiemowmde: I'm not a fan of code that's split up in so many trivial methods. But this patch does not introduces this coding style, so no reason to block it.

src/Deserializers/EntityDeserializer.php
```diff
@@ -113,24 +114,46 @@ private function setAliasesFromSerialization( array $serialization, Entity $enti
...
+			);
+		}
+	}
+
+	private function assertRequestedAndActualLanguageMatch( $serialization, $requestedLanguage ) {
+		if ( $serialization['language'] !== $requestedLanguage ) {
+			throw new DeserializationException(
+				'Deserialization of a value of the attribute language (actual)'
+					.' that is not matching the language key (requested) is not supported: '
+					.$serialization['language'].' !== '.$requestedLanguage
```
thiemowmde: Spaces.

src/Deserializers/EntityDeserializer.php
```diff
@@ -99,8 +100,8 @@ private function setDescriptionsFromSerialization( array $serialization, Entity
 	private function deserializeValuePerLanguageSerialization( array $serialization ) {
 		$array = array();
 
-		foreach( $serialization as $valueSerialization ) {
-			$this->assertIsValidValueSerialization( $valueSerialization );
+		foreach( $serialization as $requestedLanguage => $valueSerialization ) {
```
thiemowmde: For the record: This change makes the deserialization fail if the serialization does not have array keys. This is by design, as far as I understand.

src/Deserializers/EntityDeserializer.php
```diff
@@ -113,24 +114,46 @@ private function setAliasesFromSerialization( array $serialization, Entity $enti
...
 		}
 	}
 
-	private function assertIsValidValueSerialization( $serialization ) {
+	private function assertNotAttribute( array $array, $key ) {
+		if ( array_key_exists( $key, $array ) ) {
+			throw new InvalidAttributeException(
+				$key,
+				$array[$key],
+				'Deserialization of attribute '.$key.' not supported.'
```
thiemowmde: Spaces missing.

src/Serializers/FingerprintSerializer.php
thiemowmde: Ok.

src/Serializers/FingerprintSerializer.php
```diff
@@ -73,22 +76,30 @@ private function serializeValuePerLanguageArray( array $array ) {
...
+
 	public function addAliasesToSerialization( Entity $entity, array &$serialization ) {
-		$aliases = $entity->getAllAliases();
+		$aliases = $entity->getFingerprint()->getAliasGroups();
 
 		$serialization['aliases'] = $this->serializeValuesPerLanguageArray( $aliases );
 	}
 
-	private function serializeValuesPerLanguageArray( $array ) {
+	private function serializeValuesPerLanguageArray( AliasGroupList $aliases ) {
```
thiemowmde: Method name needs to change now.

src/Serializers/FingerprintSerializer.php
```diff
@@ -97,4 +108,19 @@ private function serializeValuesPerLanguageArray( $array ) {
...
+		foreach( $aliasGroup->getAliases() as $value ) {
+			$result = array(
+				'language' => $language,
+				'value' => $value
+			);
+			if ($aliasGroup instanceof AliasGroupFallback) {
+				$result['language'] = $aliasGroup->getActualLanguageCode();
+				$result['source'] = $aliasGroup->getSourceLanguageCode();
+			}
+			$serialization[$language][] = $result;
```
thiemowmde: I find this method (and the same for terms above) hard to read. I will try to refactor this in a later patch.

tests/unit/Deserializers/EntityDeserializerTest.php
```diff
@@ -212,7 +212,43 @@ public function testInvalidSerialization( $serialization ) {
 
 	public function invalidDeserializationProvider() {
 		return array(
-			array(
+			'label with integer language code' => array(
+				array(
+					'type' => 'item',
+					'labels' => array(
+						8 => array(
+							'language' => 8,
```
thiemowmde: Please add a test with
```php
'labels' => array(
array(
'language' => 'en',
```

tests/unit/Deserializers/EntityDeserializerTest.php
```diff
@@ -212,7 +212,43 @@ public function testInvalidSerialization( $serialization ) {
...
+				),
+			),
+			'alias with interger language code' => array(
+				array(
+					'type' => 'item',
+					'aliases' => array(
+						8 =>
+							array(
+								array(
+									'language' =>  8,
```
thiemowmde: As above, please add a test with no array key (defaults to int 0) and a valid language code.

src/Deserializers/EntityDeserializer.php
JanZerebecki: Done.

src/Deserializers/EntityDeserializer.php
JanZerebecki: Done.

src/Serializers/FingerprintSerializer.php
JanZerebecki: Done.

tests/unit/Deserializers/EntityDeserializerTest.php
JanZerebecki: Done.

tests/unit/Deserializers/EntityDeserializerTest.php
JanZerebecki: Done.

src/Serializers/FingerprintSerializer.php
```diff
@@ -73,22 +76,30 @@ private function serializeValuePerLanguageArray( array $array ) {
...
 
+	private function serializeTerm( Term $term, array &$serialization ) {
+		$language = $term->getLanguageCode();
+		$result = array(
+			'language' => $language,
+			'value' => $term->getText(),
+		);
+		if ( $term instanceof TermFallback ) {
+			$result['language'] = $term->getActualLanguageCode();
+			$result['source'] = $term->getSourceLanguageCode();
```
thiemowmde: So this is serialized even if it is null? Does the unserializer require this or is it flexible enough to not fail if this is missing?

src/Deserializers/EntityDeserializer.php
```diff
@@ -113,24 +114,46 @@ private function setAliasesFromSerialization( array $serialization, Entity $enti
...
+					. ' that is not matching the language key (requested) is not supported: '
+					. $serialization['language'] . ' !== ' . $requestedLanguage
+			);
+		}
+	}
+
+	private function assertIsValidValueSerialization( $serialization, $requestedLanguage ) {
 		$this->requireAttribute( $serialization, 'language' );
 		$this->requireAttribute( $serialization, 'value' );
+		$this->assertNotAttribute( $serialization, 'source' );
```
thiemowmde: So no user of this component will ever be allowed to deserialize serializations that contain fallback terms? This idea sounds a bit odd to me, if put that way. Is there an other solution?

JanZerebecki: Not until we implement deserialization and agree on a way to prevent TermFallback from accidentally being written into the DB.

PR #101, opened 2014-12-10
Precise type hints in SerializerFactory

thiemowmde:
Simple reasoning: Why should a `newReferenceSerializer` method not return a `ReferenceSerializer`?


thiemowmde:
It's a year now since I tried to make the mess these interfaces create less messy. What a pity.

PR #111, opened 2015-02-17
Drop dead code from SnakDeserializerTest

thiemowmde:


thiemowmde:
Already submitted as #108. Please merge so my PHPStorm stops complaining.

PR #114, opened 2015-03-17
Fix broken tests

JeroenDeDauw:


JeroenDeDauw:
Fixed the TravisCI build

> 

Not animooted. -1

thiemowmde:
What's wrong with Scrutinizer?

JeroenDeDauw:
It failed randomly. I amended the commit message and now it passes fine.

tests/unit/Serializers/ClaimSerializerTest.php
```diff
@@ -190,7 +190,7 @@ public function serializationProvider() {
 
 		$claim = new Statement( new Claim( new PropertyNoValueSnak( 42 ) ) );
 		$claim->setReferences( new ReferenceList( array(
-			new Reference()
+			new Reference( array( new PropertyNoValueSnak( 1 ) ) )
```
thiemowmde: This test case does not make sense any more. The `references > snaks` array below is empty. How is that possible?

tests/unit/Serializers/ReferenceListSerializerTest.php
```diff
@@ -65,10 +65,15 @@ public function serializationProvider() {
 					array(
 						'hash' => 'da39a3ee5e6b4b0d3255bfef95601890afd80709',
 						'snaks' => array()
+					),
+					array(
+						'hash' => 'da39a3ee5e6b4b0d3255bfef95601890afd80709',
+						'snaks' => array()
```
thiemowmde: Same as above. How is it possible that a Reference with 1 snak is serialized as an empty array?

thiemowmde:
-1, see the comments above.

tests/unit/Serializers/ClaimSerializerTest.php
JeroenDeDauw: This is a unit test for the ClaimSerializer, so it does not use the ReferenceSerializer. Instead, a simple fake is used.

PR #116, opened 2015-04-15
Uniform code and doc style

thiemowmde:
* Add function some level documentation.
* Logically `@throw` comes before `@return`.
* Add spaces in `if (` and such.

### DIFF
--- a/src/Deserializers/ClaimsDeserializer.php
+++ b/src/Deserializers/ClaimsDeserializer.php
@@ -31,8 +31,8 @@ class ClaimsDeserializer implements Deserializer {
 	 *
 	 * @param array $serialization
 	 *
-	 * @return Claims
 	 * @throws DeserializationException
+	 * @return Claims
 	 */
 	public function deserialize( $serialization ) {
 		$this->assertHasGoodFormat( $serialization );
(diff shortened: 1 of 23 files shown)


src/Deserializers/EntityDeserializer.php
```diff
@@ -59,6 +59,11 @@ public function deserialize( $serialization ) {
 	 */
 	protected abstract function getPartiallyDeserialized( array $serialization );
 
+	/**
+	 * @param array $serialization
+	 *
+	 * @return Entity
```
JeroenDeDauw: EntityDocument please

JeroenDeDauw:
+1. If my comment is addressed, +2. Willing to make this change myself.

src/Deserializers/EntityDeserializer.php
thiemowmde: I tried but it's not that easy. The whole class depends on Entity. Please merge this and try to phase out Entity in an other patch.

JeroenDeDauw: What do you mean with it not being that easy? What problem occurs when using this type hint?

thiemowmde: It's not true. A few lines below the code returns an `Entity`.

JeroenDeDauw: Then it is true, since Entity is a subtype of EntityDocument.

thiemowmde: Ok, both is true. But this class is not called `EntityDocumentDeserializer`.

JeroenDeDauw: Sure. So will you change it? Or let me change it? If not, that's fine, though then I'm sticking with my +1.

thiemowmde: Change what? The class name?

Benestar:
@thiemowmde Are you going to rebase this or should it be closed?

thiemowmde:
Why should a 1 line conflict make a 100+ lines patch obsolete?

PR #119, opened 2015-04-27
Replace Snaks interface with SnakList

thiemowmde:
~~This requires https://github.com/wmde/WikibaseDataModel/pull/467.~~


Benestar:
Shouldn't we rename the `SnaksSerializer` to `SnakListSerializer` then as the `Snaks` interface is to be removed and the class will only be able to handle `SnakList` objects?

JeroenDeDauw:
That'd be nice yes, not something required for this commit though

Benestar:
Done within #132

thiemowmde:
Not 100%.

Issue #82, opened 2014-09-17
Replace ClaimsSerializer by StatementListSerializer


Benestar:
This is needed for #80

PR #115, opened 2015-03-23
Switch from DataModel 2.x to 3.x

JeroenDeDauw:

PR #120, opened 2015-04-30
Created StatementListSerializer and StatementListDeserializer

Benestar:
This is a step towards #82 and to get rid of the deprecated `Claims` class.

In follow up patches I will replace `ClaimsSerializer` and `ClaimsDeserializer` with these classes. However, there are even more changes needed in `EntityDeserializer` to make it really sane.

<s>**Note:** missing tests (tbd)</s>

### DIFF
--- a/src/Deserializers/StatementListDeserializer.php
+++ b/src/Deserializers/StatementListDeserializer.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Wikibase\DataModel\Deserializers;
+
+use Deserializers\Deserializer;
+use Deserializers\Exceptions\DeserializationException;
+use Wikibase\DataModel\Statement\StatementList;
+
+/**
+ * Package private
+ *
+ * @licence GNU GPL v2+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+class StatementListDeserializer implements Deserializer {
+
+	/**
+	 * @var Deserializer
+	 */
+	private $statementDeserializer;
+
+	/**
+	 * @param Deserializer $statementDeserializer
+	 */
+	public function __construct( Deserializer $statementDeserializer ) {
+		$this->statementDeserializer = $statementDeserializer;
+	}
+
+	/**
+	 * @see Deserializer::deserialize
+	 *
+	 * @param mixed $serialization
+	 *
+	 * @return object
+	 * @throws DeserializationException
+	 */
+	public function deserialize( $serialization ) {
+		$this->assertHasGoodFormat( $serialization );
+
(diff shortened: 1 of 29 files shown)


tests/unit/Serializers/StatementListSerializerTest.php
```diff
@@ -0,0 +1,122 @@
...
+				),
+				new StatementList( array(
+					$statement
+				) )
+			),
+		);
+	}
+
+	public function testStatementListSerializerWithOptionObjectsForMaps() {
+		$statement = new Statement( new Claim( new PropertyNoValueSnak( 42 ) ) );
```
JeroenDeDauw: This is going to break when we switch to DataModel 3.x. Which is why I think we should first do #115

src/Deserializers/StatementListDeserializer.php
```diff
@@ -0,0 +1,67 @@
...
+	public function deserialize( $serialization ) {
+		$this->assertHasGoodFormat( $serialization );
+
+		return $this->getDeserialized( $serialization );
+	}
+
+	private function getDeserialized( array $serialization ) {
+		$statementList = new StatementList();
+
+		foreach( $serialization as $statementArray ) {
```
JeroenDeDauw: I spend some time wondering why we have a list of lists here. Having some kind of hint that these are grouped by property id would be nice. Which also makes me wonder if we should just accept statements with a property id different than that of the list they are in. Though that is not something to change here if this just re-implements the current code.

JeroenDeDauw:
Generally looks good

tests/unit/Serializers/StatementListSerializerTest.php
Benestar: Sure, however, it can also be fixed in #115 thus imo the order of both patches doesn't really matter.

src/Deserializers/StatementListDeserializer.php
Benestar: It indeed just reflects the current serialization format. I'm not sure if we should enforce statements grouped by property id in the serializatoin though if it is not done in the DataModel. This is a more general question but we should stay consistent and either do this grouping in Items/Properties as well or don't force it in the serialization.

JeroenDeDauw:
Needs rebase

tests/unit/Serializers/StatementListSerializerTest.php
JeroenDeDauw: It's more work though :)

Benestar: I can adjust your PR if you prefer to safe you time ;-)

PR #121, opened 2015-04-30
Use StatementListSerializer where appropriate

Benestar:
This replaces the `claims` key with a `statements` key in the serialization which is a breaking change. We might consider to use `claims` in future as well for b/c but imo it would be cleaner to change the serialization key as well when we discard the whole concept of claims.

See #82 and #120

### DIFF
--- a/src/DeserializerFactory.php
+++ b/src/DeserializerFactory.php
@@ -15,6 +15,7 @@ use Wikibase\DataModel\Deserializers\ReferenceListDeserializer;
 use Wikibase\DataModel\Deserializers\SiteLinkDeserializer;
 use Wikibase\DataModel\Deserializers\SnakDeserializer;
 use Wikibase\DataModel\Deserializers\SnakListDeserializer;
+use Wikibase\DataModel\Deserializers\StatementListDeserializer;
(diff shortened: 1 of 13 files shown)


JeroenDeDauw:
We might or might not want to make the breaking change to the serialization format. That question really transcends this component though. And it's probably part of the same discussion as changing API module names and parameters in Wikibase Repository that contain "claim". I suggest to raise this question elsewhere.

If we where to merge this now, then we can no longer switch Wikibase Repository over to use DataModel 3.x without also focusing this format change on it. That's not a good situation. Let's first make this component work with DataModel 3.x, make a release, and then make the format change, assuming we'd want to go ahead with it then.

Benestar:
Ok, so I will replace `statements` with `claims` and leave this discussion open for now.

Benestar:
This is no longer a breaking change and is thus useful in Wikibase Repository as it is.

JeroenDeDauw:
Looks good after a quick glance. Can you put this into a single decently named commit? I'd rather not have the breaking change be part of the actual git history.

Benestar:
Sure, I can combine the commits into one so that it does not pollute the log.

PR #124, opened 2015-05-01
Remove ClaimsSerializer and ClaimsDeserializer completely

Benestar:
They are not used anymore in this component and replaced completely by `StatementListSerializer` and `StatementListDeserializer`.


JeroenDeDauw:
As far as I'm aware, there have no breaking changes on master yet, so we can still do a 1.4 release, rather than a 2.0 one. And making a non-breaking release is nicer than making a breaking one. So I'm wondering if we really need to remove those `newClaimsSerializer` and `newClaimsDeserializer` factory methods, or if we cannot simply deprecate them till a 2.0 release. I suspect that still supporting this is easy, by doing something like this in the appropriate implementation code:

```php
if ( $statements instanceof Claims ) {
    $statements = new StatementList( $statements );
}
```

Benestar:
Yes, perhaps that is a better idea. Removing those factory methods can be done when we finally have a new major release. They don't hurt anyways.

JeroenDeDauw:
It's not just the factory functions btw - also actually supporting the use of `Claims`, which might still be needed in places.

JeroenDeDauw:
@Benestar ping

Benestar:
We can make the ClaimsSerializer/Deserializer internally using StatementListSerializer/Deserializer. However, removing the classes completely should only be done when Claims is removed in the DataModel, so on hold for now.

JeroenDeDauw:
How about making a commit with just the added integration test and then closing the rest?

Benestar:
I will rewrite this commit when I have some time for it ;-)

PR #127, opened 2015-05-04
More specific type hints in Serializers

thiemowmde:
All these parameters must be of this type, otherwise the call throws an exception. It's guaranteed to do so. This is part of the public interface. So this has nothing to do with "package private" or something.

### DIFF
--- a/src/Serializers/ClaimSerializer.php
+++ b/src/Serializers/ClaimSerializer.php
@@ -65,7 +65,7 @@ class ClaimSerializer implements DispatchableSerializer {
 	/**
 	 * @see Serializer::serialize
 	 *
-	 * @param mixed $object
+	 * @param Claim $object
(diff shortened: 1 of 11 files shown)


Benestar:
I also wondered about this typehints but I think the `mixed` typehint makes sense because the `Serializer` interface requires that any object type can be passed to that methods. Throwing an exception does not make that type invalid.

thiemowmde:
@Benestar Sorry, but I can't accept this as an argument. The method dies immediately if the parameter does not have the correct type. This must be documented. This is consistent to countless other places in our code base, for example:
```php
function example( $var ) {
	if ( !is_int( $var ) ) {
		throw new InvalidArgumentException();
```
We document this with `@param int $var`, not with `@param mixed $var`.

PR #128, opened 2015-05-04
More specific type hints in Deserializers

thiemowmde:
All these parameters and return types must be of this type, otherwise the call throws an exception. It's guaranteed to do so. This is part of the public interface. So this has nothing to do with "package private" or something.

### DIFF
--- a/src/Deserializers/ClaimDeserializer.php
+++ b/src/Deserializers/ClaimDeserializer.php
@@ -69,9 +69,9 @@ class ClaimDeserializer implements DispatchableDeserializer {
 	/**
 	 * @see Deserializer::deserialize
 	 *
-	 * @param mixed $serialization
+	 * @param array $serialization
 	 *
-	 * @return object
+	 * @return Claim|Statement
 	 * @throws DeserializationException
 	 */
 	public function deserialize( $serialization ) {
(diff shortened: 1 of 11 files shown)


Benestar:
See https://github.com/wmde/WikibaseDataModelSerialization/pull/127#issuecomment-98791105

This does not represent the guarantee provided by the `Deserializer` interface but still reflects the guarantees of the actual implementations. I'm merging this as the other patch got merged as well, for consistency.

src/Deserializers/ClaimDeserializer.php
```diff
@@ -69,9 +69,9 @@ private function hasCorrectType( $serialization ) {
 	/**
 	 * @see Deserializer::deserialize
 	 *
-	 * @param mixed $serialization
+	 * @param array $serialization
 	 *
-	 * @return object
+	 * @return Claim|Statement
```
Benestar: Uh, didn't see that. `Claim` should last here because `Statement` is still a subclass of `Claim` in the version of DM we depend here.

thiemowmde:
See https://github.com/wmde/WikibaseDataModelSerialization/pull/127#issuecomment-99005955 as well as https://github.com/wmde/WikibaseDataModelSerialization/pull/130#issuecomment-99003203.

PR #130, opened 2015-05-04
Remove redundant typehint

Benestar:
In DataModel 2.6, `Statement` is still a subclass of `Claim` and thus doesn't need to be mentioned here.

### DIFF
--- a/src/Deserializers/ClaimDeserializer.php
+++ b/src/Deserializers/ClaimDeserializer.php
@@ -71,7 +71,7 @@ class ClaimDeserializer implements DispatchableDeserializer {
 	 *
 	 * @param array $serialization
 	 *
-	 * @return Claim|Statement
+	 * @return Claim
 	 * @throws DeserializationException
 	 */
 	public function deserialize( $serialization ) {


thiemowmde:
But it's not wrong and allows for easier, more obvious migration when we switch to DataModel 3.0. Can you please close this?

Benestar:
I do not see any reason why this allows for easier, more obvious migration. When migrating, simply replacing Claim by Statement should work in most places, also here. So I still think this typehint is redundant and has no benefit.

thiemowmde:
The method is guaranteed to either return a `Claim` or a `Statement` object. Nothing else, especially no other `Claim` subclass. That's exactly what the documentation says and what I wanted it to say. It's not wrong and does not need to be fixed. Please close this.

Benestar:
A `Statement` object actually is a `Claim` object. However, discussing this is idle as it won't matter when we have done the switch to DM 3.0 (#115).

thiemowmde:
So what? This does not mean it can not be documented.

PR #131, opened 2015-05-04
Remove EntityDeserializer abstract base class

Benestar:
This class was never exposed in public but only used for code sharing.
All relevant code has been moved out into separate deserializers so this
can now be safely deleted.

Remember that this is not a breaking change as EntityDeserializer was
not a public class but only an implementation detail.


JeroenDeDauw:
Yay for narrow interfaces \o/

PR #132, opened 2015-06-03
Add compatibility with DataModel 3.0

Benestar:
This patch adapts the changes made in DataModel 3.0 and makes WikibaseDataModelSerialization ready for that release. @JanZerebecki


JeroenDeDauw:
So this is basically what #115 already does?

Benestar:
Heh, didn't see that -.- also, your patch seems to be a bit older and the tests fail there :S

JeroenDeDauw:
Well yeah, this is one of the things that happen when we just leave such stuff sit there

addshore:
All looks good!
I guess we can close #115 now?

PR #133, opened 2015-06-08
1.4 release

Benestar:


JeroenDeDauw:
Note that for just modifying the docs files you can skip the build using by placing `[skip ci]` in the commit description.

I'd like to have a closer look at this thing in my IDE before release, to check if there is not something else that really makes sense to put in here. Of course if work elsewhere is made significantly easier by already releasing, then perhaps we should just do that, since additional changes can go into a second release soon after without causing issues.

JeroenDeDauw:
Let's finish up #115 first?

Benestar:
@JeroenDeDauw did you find any other issues in your IDE that you want to have in this release? Otherwise I'd say we can do this.

JeroenDeDauw:
Go for it. Several things I want to do, but they can all wait.

RELEASE-NOTES.md
```diff
@@ -1,14 +1,14 @@
 # Wikibase DataModel Serialization release notes
 
-## 1.4.0 (dev)
+## 1.4.0 (2015-06-08)
 
 * Added support for showing the component version when loaded via MediaWiki
 * Compatibility with DataModel 3.x added
 * DataModel 3.x is now required
```
thiemowmde: Aren't these two lines redundant?

Benestar: Per https://github.com/wmde/WikibaseDataModelSerialization/blob/master/RELEASE-NOTES.md#110-2014-09-02

PR #135, opened 2015-06-08
Add missing release notes and @since tags

Benestar:
If I didn't miss something those things also got added in 1.4.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -11,7 +11,6 @@
 * Renamed `DeserializerFactory::newSnaksDeserializer` to `DeserializerFactory::newSnakListDeserializer`, leaving a b/c alias
 * Added `SerializerFactory::newStatementListSerializer`
 * Added `DeserializerFactory::newStatementListDeserializer`
-* Added `DeserializerFactory::newFingerprintDeserializer`
 
 ## 1.3.0 (2015-01-28)
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,6 +9,9 @@
 * Renamed `SerializerFactory::newSnaksSerializer` to `SerializerFactory::newSnakListSerializer`, leaving a b/c alias
 * Renamed `DeserializerFactory::newClaimDeserializer` to `DeserializerFactory::newStatementDeserializer`, leaving a b/c alias
 * Renamed `DeserializerFactory::newSnaksDeserializer` to `DeserializerFactory::newSnakListDeserializer`, leaving a b/c alias
+* Added `SerializerFactory::newStatementListSerializer`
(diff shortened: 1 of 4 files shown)


RELEASE-NOTES.md
```diff
@@ -9,6 +9,9 @@
 * Renamed `SerializerFactory::newSnaksSerializer` to `SerializerFactory::newSnakListSerializer`, leaving a b/c alias
 * Renamed `DeserializerFactory::newClaimDeserializer` to `DeserializerFactory::newStatementDeserializer`, leaving a b/c alias
 * Renamed `DeserializerFactory::newSnaksDeserializer` to `DeserializerFactory::newSnakListDeserializer`, leaving a b/c alias
+* Added `SerializerFactory::newStatementListSerializer`
+* Added `DeserializerFactory::newStatementListDeserializer`
+* Added `DeserializerFactory::newFingerprintDeserializer`
```
thiemowmde: Hm, why is there no FingerprintSerializer?

Benestar: There is, we just don't have it publicly exposed in a Factory.

thiemowmde: To be more precise, why is the deserializer public via the factory but the serializer is not?

thiemowmde: I think both should not be exposed. There currently is no need for it as far as I can see.

src/DeserializerFactory.php
```diff
@@ -55,8 +55,8 @@ public function __construct( Deserializer $dataValueDeserializer, EntityIdParser
 	 */
 	public function newEntityDeserializer() {
 		return new DispatchingDeserializer( array(
-			new ItemDeserializer( $this->newEntityIdDeserializer(), $this->newFingerprintDeserializer(), $this->newStatementListDeserializer(), $this->newSiteLinkDeserializer() ),
-			new PropertyDeserializer( $this->newEntityIdDeserializer(), $this->newFingerprintDeserializer(), $this->newStatementListDeserializer() )
+			new ItemDeserializer( $this->newEntityIdDeserializer(), new FingerprintDeserializer(), $this->newStatementListDeserializer(), $this->newSiteLinkDeserializer() ),
+			new PropertyDeserializer( $this->newEntityIdDeserializer(), new FingerprintDeserializer(), $this->newStatementListDeserializer() )
```
thiemowmde: Please reuse the same object.

thiemowmde:
I checked and I think everything is covered now and ready to be released.

src/DeserializerFactory.php
Benestar: Fixed. This issue already existed before though.

thiemowmde: Cool, thanks.

PR #136, opened 2015-06-08
Stop using deprecated method in ItemSerializer

JeroenDeDauw:


src/Serializers/ItemSerializer.php
```diff
@@ -102,13 +102,12 @@ private function addStatementListToSerialization( Item $item, array &$serializat
 	}
 
 	private function addSiteLinksToSerialization( Item $item, array &$serialization ) {
-		$siteLinks = $item->getSiteLinks();
-
 		$serialization['sitelinks'] = array();
 
-		foreach( $siteLinks as $siteLink ) {
+		foreach( $item->getSiteLinkList()->toArray() as $siteLink ) {
```
Benestar: `SiteLinkList` implements `IteratorAggregate` so you don't have to call `toArray` here, right?

thiemowmde: @JeroenDeDauw does this to avoid the line `/** @var SiteLink $siteLink */`. Both is valid. Personally I like the `@var` solution more.

thiemowmde:
@JeroenDeDauw, needs a rebase now. Would +2 otherwise.

JeroenDeDauw:
Let's just start using PHP7 of Hack... or Python :)

src/Serializers/ItemSerializer.php
```diff
@@ -107,13 +107,12 @@ private function addStatementListToSerialization( Item $item, array &$serializat
 	}
 
 	private function addSiteLinksToSerialization( Item $item, array &$serialization ) {
-		$siteLinks = $item->getSiteLinks();
-
 		$serialization['sitelinks'] = array();
-
-		foreach ( $siteLinks as $siteLink ) {
+
```
thiemowmde: Tabs added?

thiemowmde:
+2 when tests pass.

PR #137, opened 2015-06-08
Decrease binding to Entity and derivatives

JeroenDeDauw:


src/Serializers/PropertySerializer.php
```diff
@@ -72,7 +72,12 @@ private function getSerialized( Property $entity ) {
 			'datatype' => $entity->getDataTypeId(),
 		);
 
-		$this->fingerprintSerializer->addBasicsToSerialization( $entity, $serialization );
```
Benestar: $entity -> $property

src/Serializers/FingerprintSerializer.php
```diff
@@ -48,15 +47,11 @@ public function addIdToSerialization( Entity $entity, array &$serialization ) {
 		$serialization['id'] = $id->getSerialization();
 	}
 
-	public function addLabelsToSerialization( Entity $entity, array &$serialization ) {
-		$labels = $entity->getFingerprint()->getLabels();
-
+	public function addLabelsToSerialization( TermList $labels, array &$serialization ) {
```
thiemowmde: Many methods here can be private. Not part of this patch, just wanted to point this out.

src/Serializers/PropertySerializer.php
thiemowmde: Correct, but not part of this patch.

tests/unit/Serializers/FingerprintSerializerTest.php
```diff
@@ -312,41 +315,47 @@ public function serializationProvider() {
...
+		$descriptions->setTextForLanguage( 'en', 'A Nyan Cat' );
 
 		$result = array();
 
-		$descriptions = new stdClass();
-		$descriptions->en = array(
-			'language' => 'en',
-			'value' => 'A Nyan Cat'
+		$expectedSerialization = array(
+			'descriptions' => (object)array(
```
thiemowmde: Uh, nice. I like that.

src/Serializers/FingerprintSerializer.php
JeroenDeDauw: Yeah... I'm quite unhappy about what that code looked like, this commit only fixing some of it. Too bad it got through review somehow

tests/unit/Serializers/FingerprintSerializerTest.php
Benestar: `use stdClass;` can be removed then, right?

thiemowmde: Already done in #140.

Benestar: :+1:

PR #138, opened 2015-06-08
Use constants to avoid: new Foo(true)

JeroenDeDauw:


JeroenDeDauw:
This commit is made on top of #137, which you probably want to review separately.

thiemowmde:
Besides my comment (see above) the second commit (6e615571a84d5d722839cf6174602e8daf153b3d) is fine for me. This is all package private anyway. I will review the first commit in #137.

Note that @brightbyte prefers a slightly other style when coming across this problem: `new FingerprintSerializer( 'useObjectsForMaps' )` with the option not being a bool but a string. Personally I do not have a preference.

JeroenDeDauw:
If you put it as a sting, then IDEs and other tools will not recognize it, and typos will be a lot less obvious.

thiemowmde:
@JeroenDeDauw, can you rebase this please? Would be much easier to review and merge then.

JeroenDeDauw:
I'm somewhat confused by the first commit not disappearing from the PR as it has been merged already... will rebase

src/SerializerFactory.php
```diff
@@ -69,7 +69,8 @@ private function shouldUseObjectsForMaps() {
 	 * @return Serializer
 	 */
 	public function newEntitySerializer() {
-		$fingerprintSerializer = new FingerprintSerializer( $this->shouldUseObjectsForMaps() );
+		$options = $this->shouldUseObjectsForMaps() ? FingerprintSerializer::USE_OBJECTS_FOR_MAPS : FingerprintSerializer::USE_ARRAYS_FOR_MAPS;
```
thiemowmde: Please rename, these are no "options".

JeroenDeDauw: Any suggestion what to rename it to?

thiemowmde:
```php
$useObjectsForMaps = $this->shouldUseObjectsForMaps()
	? FingerprintSerializer::USE_OBJECTS_FOR_MAPS
	: FingerprintSerializer::USE_ARRAYS_FOR_MAPS;
```

JeroenDeDauw: That is a boolean name. The fact that the constants use a boolean is a detail that should now be hidden. It is just the "map serialization option" now. So how about `$mapSerializationOption` then?

thiemowmde: `$option` is fine, just want to avoid confusion with `SerializationOptions`.

Benestar: Imo we should make this a bitwise integer option like in `SerializerFactory` and we should also reuse the constants from there.

JeroenDeDauw: That'd make the serializers depend on a factory, which is not nice, even when ignoring the cyclic dependency. We could create a dedicated enum. But perhaps that ends up overdoing it?

Again, Python does not have this issue at all... :|

thiemowmde: Please do not use C style bit fields. Either constants or short strings. The bool constants in this patch are perfectly fine. All this is package private anyway, isn't it?

JeroenDeDauw: > All this is package private anyway, isn't it?

Yes

thiemowmde:
+2 from my side. @JeroenDeDauw, please have a look if the commit 2745b0528486c0c863e2c512cd671434e90ed1f7 I added is ok.

PR #145, opened 2015-06-24
Deprecate the Claims (de)serializer factory methods

JeroenDeDauw:


thiemowmde:
Does this really need a +0.1.0 "new feature" step for the version number?

Benestar:
Yes, per http://semver.org/:

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. **It MUST be incremented if any public API functionality is marked as deprecated.**

thiemowmde:
Good to know, thanks.

PR #150, opened 2015-06-30
Fixes addshore

Benestar:
Remove Fingerprint[De]Serializer and fix lots of stuff

Fixes @addshore


JeroenDeDauw:
"Fixes addshore", now forever in the git log :)

JeroenDeDauw:
Actually, you want to do a git log and go the the very beginning of the log... :)

Warning: you might also spontaneously faceplam when you consider how old this component already is and that the lib code is still not gone

PR #152, opened 2015-07-09
Add Serialization option for snak hashes

addshore:
In the current Wikibase Lib serializers there is
an option to serialize snaks without a hash.

This is due to the fact that the hashes are not all
that usefull in the API output.

To avoid adding these hashes to the serialization
this option is needed

### DIFF
--- a/src/SerializerFactory.php
+++ b/src/SerializerFactory.php
@@ -73,7 +73,7 @@ class SerializerFactory {
 	 * @return bool
 	 */
 	private function shouldSerializeSnaksWithHash() {
-		return !(bool)( $this->options & self::OPTION_OBJECTS_FOR_MAPS );
+		return !(bool)( $this->options & self::OPTION_SERIALIZE_SNAKS_WITHOUT_HASH );
 	}
 
 	/**
--- a/src/SerializerFactory.php
+++ b/src/SerializerFactory.php
@@ -29,11 +29,13 @@ use Wikibase\DataModel\Serializers\TypedSnakSerializer;
  * @licence GNU GPL v2+
  * @author Thomas Pellissier Tanon
  * @author Bene* < benestar.wikimedia@gmail.com >
+ * @author Adam Shorland
  */
 class SerializerFactory {
 
 	const OPTION_DEFAULT = 0;
 	const OPTION_OBJECTS_FOR_MAPS = 1;
+	const OPTION_SERIALIZE_SNAKS_WITHOUT_HASH = 2;
 
 	/**
 	 * @var int
@@ -65,6 +67,13 @@ class SerializerFactory {
 	 */
 	private function shouldUseObjectsForMaps() {
 		return (bool)( $this->options & self::OPTION_OBJECTS_FOR_MAPS );
+	}
+
+	/**
+	 * @return bool
+	 */
+	private function shouldSerializeSnaksWithHash() {
+		return !(bool)( $this->options & self::OPTION_OBJECTS_FOR_MAPS );
 	}
 
 	/**
@@ -192,7 +201,10 @@ class SerializerFactory {
 	 * @return Serializer
 	 */
 	public function newSnakSerializer() {
-		return new SnakSerializer( $this->dataValueSerializer );
+		return new SnakSerializer(
+			$this->dataValueSerializer,
+			$this->shouldSerializeSnaksWithHash()
+		);
 	}
 
 	/**
--- a/src/SerializerFactory.php
+++ b/src/SerializerFactory.php
@@ -34,8 +34,12 @@ use Wikibase\DataModel\Serializers\TypedSnakSerializer;
 class SerializerFactory {
 
 	const OPTION_DEFAULT = 0;
+	/** @since 1.2.0 */
 	const OPTION_OBJECTS_FOR_MAPS = 1;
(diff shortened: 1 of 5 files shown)


src/SerializerFactory.php
```diff
@@ -67,6 +69,13 @@ private function shouldUseObjectsForMaps() {
 	}
 
 	/**
+	 * @return bool
+	 */
+	private function shouldSerializeSnaksWithHash() {
+		return !(bool)( $this->options & self::OPTION_OBJECTS_FOR_MAPS );
```
thiemowmde: Copypasto!

src/Serializers/SnakSerializer.php
```diff
@@ -23,10 +24,17 @@ class SnakSerializer implements DispatchableSerializer {
...
+	 * @var bool
+	 */
+	private $serializeSnaksWithHash;
+
+	/**
 	 * @param Serializer $dataValueSerializer
+	 * @param bool $serializeSnaksWithHash
 	 */
-	public function __construct( Serializer $dataValueSerializer ) {
+	public function __construct( Serializer $dataValueSerializer, $serializeSnaksWithHash = true ) {
```
thiemowmde: The variable name is a bit complicated. We already know we are in the SnakSerializer. `$serializeHash` should be enough.

JeroenDeDauw: Think that name is not 100% correct grammar wise. I'd go for either `serializeWithHash` or something like `includeHash`

src/SerializerFactory.php
JeroenDeDauw: ? I'm no seeing what is wrong. Is there a typo?

JeroenDeDauw:
Is this an option that really should be global to the factory? Do we not have use cases where we want to serialize data that includes snaks and only want some of them with a hash?

src/Serializers/SnakSerializer.php
Benestar: I agree and like `$includeHash` most

src/SerializerFactory.php
Benestar: I'm not blocking the view - I am the view.

Jeroen doesn't see the typo for all the kittens on his screen.

JeroenDeDauw: Indeed. Seeing the cat. Not seeing any typo :)

thiemowmde: OPTION_OBJECTS_FOR_MAPS is the *other* option.

JeroenDeDauw: Oh... well, the cat is more interesting anyway. puh!

Anyway, the test passed, so all must therefore be good.

nope-still-good

addshore:
Currently I cant find such a use case. One might pop up in the future though..
Multiple options to cover snaks hashes inside of different elements? (If we continue to do options in the same way)

Benestar:
Well, "You aren't gonna need it" is the appropriate quotation in this case ;-) Let's think about this if we encounter such a situation.

addshore:
Hah, well, not gonna need it at this stage.

JeroenDeDauw:
> Multiple options to cover snaks hashes inside of different elements?

I'd give the option in the factory method then, and just name the method appropriately (2 versions).

> Well, "You aren't gonna need it" is the appropriate quotation in this case ;-)

Not really. This is not about adding support for something we do not need yet. There is no extra complexity. (In fact complexity and binding will be lower if we do the above.) It is designing in such a way that we minimize the risk of things getting messy.

addshore:
Well, we would end up needing 1 bool option in / 1 alternative method for:
 - newTypedSnakSerializer
 - newSnakSerializer
 - newSnakListSerializer
 - newReferenceSerializer
 - newReferencesSerializer
and 3 bool options / 8 alternative methods for:
 - newStatementSerializer
 - newStatementListSerializer
 - newEntitySerializer

Or the best approach I feel is have an single option which is an array snak hashes to not include, for example:
    array( 'reference', 'qualifier )

Benestar:
I still prefer the global option. Why should one want to include snak hashes in references but not in qualifiers? That simply doesn't make sense. In the api output, snak hashes aren't helpful anywhere and other places want them in both. Let's keep it simple and have a single option.

JeroenDeDauw:
Did not realize we'd need it for all these methods. Global option it is then...

addshore:
@Benestar well, the case would be that reference snak hashes have no use, qualifier snak hashes are needed by the API, statement mainsnak hashes have no use.. etc..

addshore:
Closing this for now.

Benestar:
Why closed? I think we agreed on having the option as implemented here?

addshore:
No option at all would be cleaner ;)

Benestar:
I thought we didn't want to break the api? o.O :P

addshore:
Well, extra keys wouldn't be a break, it would just add extra data / extra confusion

addshore:
So after looking at the Serialization stuff more I think I am going to re open this PR but change from 1 generic option to 3 specific options.

OPTION_SERIALIZE_MAIN_SNAKS_WITHOUT_HASH
OPTION_SERIALIZE_QUALIFER_SNAKS_WITHOUT_HASH
OPTION_SERIALIZE_REFERENCE_SNAKS_WITHOUT_HASH

PR #156, opened 2015-07-23
Remove old branch alias

addshore:


JeroenDeDauw:
Oh, not updated... Any reason why you are removing rather than updating? I've submitted an update in #158

PR #162, opened 2015-07-29
Support DerivedPropertyValueSnak in SnakSerializer

Benestar:
This is a draft how we could support derived data values in `SnakSerializer`. Another option would be to introduce an extra key like `slots` or `derived-values` where we add the extra data values. If we consider using this solution, there might be conflicts with existing keys.

### DIFF
--- a/src/Deserializers/SnakDeserializer.php
+++ b/src/Deserializers/SnakDeserializer.php
@@ -107,6 +107,8 @@ class SnakDeserializer implements DispatchableDeserializer {
 
 	private function newValueSnak( array $serialization ) {
 		$this->requireAttribute( $serialization, 'datavalue' );
+		// Do not deserialize derived value snaks
+		$this->assertNotAttribute( $serialization, 'derived-values' );
 
 		return new PropertyValueSnak(
 			$this->deserializePropertyId( $serialization['property'] ),
@@ -156,6 +158,16 @@ class SnakDeserializer implements DispatchableDeserializer {
 		if ( !array_key_exists( $attributeName, $array ) ) {
 			throw new MissingAttributeException(
 				$attributeName
+			);
+		}
+	}
+
+	private function assertNotAttribute( array $array, $key ) {
+		if ( array_key_exists( $key, $array ) ) {
+			throw new InvalidAttributeException(
+				$key,
+				$array[$key],
+				'Deserialization of attribute ' . $key . ' not supported.'
 			);
 		}
 	}
--- a/src/Serializers/SnakSerializer.php
+++ b/src/Serializers/SnakSerializer.php
@@ -6,6 +6,7 @@ use Serializers\DispatchableSerializer;
 use Serializers\Exceptions\SerializationException;
 use Serializers\Exceptions\UnsupportedObjectException;
 use Serializers\Serializer;
+use Wikibase\DataModel\Snak\DerivedPropertyValueSnak;
 use Wikibase\DataModel\Snak\PropertyValueSnak;
 use Wikibase\DataModel\Snak\Snak;
 
@@ -79,6 +80,12 @@ class SnakSerializer implements DispatchableSerializer {
 
 		if ( $snak instanceof PropertyValueSnak ) {
 			$serialization['datavalue'] = $this->dataValueSerializer->serialize( $snak->getDataValue() );
+		}
+
+		if ( $snak instanceof DerivedPropertyValueSnak ) {
+			foreach ( $snak->getDerivedDataValues() as $key => $dataValue ) {
+				$serialization[$key] = $this->dataValueSerializer->serialize( $dataValue );
+			}
 		}
 
 		return $serialization;
--- a/src/Serializers/SnakSerializer.php
+++ b/src/Serializers/SnakSerializer.php
@@ -84,7 +84,7 @@ class SnakSerializer implements DispatchableSerializer {
 
 		if ( $snak instanceof DerivedPropertyValueSnak ) {
 			foreach ( $snak->getDerivedDataValues() as $key => $dataValue ) {
-				$serialization[$key] = $this->dataValueSerializer->serialize( $dataValue );
+				$serialization['derived-values'][$key] = $this->dataValueSerializer->serialize( $dataValue );
 			}
 		}
 
--- a/src/Serializers/SnakSerializer.php
+++ b/src/Serializers/SnakSerializer.php
@@ -84,7 +84,7 @@ class SnakSerializer implements DispatchableSerializer {
 
 		if ( $snak instanceof DerivedPropertyValueSnak ) {
 			foreach ( $snak->getDerivedDataValues() as $key => $dataValue ) {
-				$serialization['derived-values'][$key] = $this->dataValueSerializer->serialize( $dataValue );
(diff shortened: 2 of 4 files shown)


RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel Serialization release notes
 
+## 1.8.1 (dev)
```
JeroenDeDauw: 1.9 as per semver.org

JeroenDeDauw:
I find it very strange to merge in a map of stuff with the main properties of the object. We are not doing this anywhere else either. Having the map under its own key, such as `derived-values` is a lot less surprising to me.

What are these keys anyway? I looked the the DataModel object, but this leaves them completely unspecified.

addshore:
I had another think about this last night, while sat on a dark beach but with an epic moon ;)

All / Most datamodel objects such as snak, sitelink could be interfaces.
There would also be a basic implementation of the interfaces ie, the objects we are currently using.
All of this would be in DataModel ofc

There would be a "Slotty" or "Derived" interface too, this interface could perhaps even be in the serialization component>
The functionality would basically be the same as that seen in the DerivedSnak thing

We would them implement a bunch of "Slotty" objects....
SlottySiteLink (would enable us to add the url into serialization easily..
SlottySnak (would allow us to add the datatype)
etc.
This would either be in Wikibase or DMSerialization.

Thoughts? (I may knock up some PRs to see what people think!

addshore:
Please take a look at https://github.com/wmde/WikibaseDataModelSerialization/pull/163 and comment..

Benestar:
Just as a side note, this would break the whole DataModel again /o\

The idea of only dealing with interfaces only is already implemented in [Wikidata Toolkit](https://github.com/Wikidata/Wikidata-Toolkit/tree/master/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel).

JeroenDeDauw:
"Slotty"

the-fuck-am-i-reading

JeroenDeDauw:
My thoughts:

* Interfaces for value objects are odd
* Making such a big breaking change requires serious justification
* Lack of explicit schema or defined structure does not remove the actual schema or structure, or the need to deal with it somewhere
* Doing instanceof checks like in the linked PR is a design smell
* Making such invasive changes the DataModel implementation like this for the need of a small subset of its users is very suspicious

addshore:
Well, I just realised that none of this slotty stuff need go near the datamodel itself, just in serialization, and everything could just be done as in the PR with a single interface and no breaking changes.

That is of course unless we prefer this approach, which I personally don't right now.
A similar thing to this is the TypedSnak, which although fully implemented in DM and DMS is not actually used in Wikibase because, well, it cant be / there is no actual use for it.

This slotty thing (yes it could probably use a better name)... could immediately be used to remove buckets of hacky code in the API. It would probably also mean this DerivedPropertyValueSnak and the TypedValue snak are both redundant.

JeroenDeDauw:
> ...DerivedPropertyValueSnak and the TypedValue snak are both...

These two are not quite the same though. TypedSnak fully defines what it is and does not derive from Snak. To me it is closer to things such as StatementByGuidMap than to things such as DerivedPropertyValueSnak, the later of which pretends to be a regular snak and has a bunch of schema-less information.

addshore:
Well we could probably rework the slot stuff to explicitly define what it accepts for each object, thus it would have a schema!

Still the slotty serialization stuff would likely be in DataModelSerializaion with the interface also in the component but then the slotty objects implemented in WIkibase?

Benestar:
Can we please discuss the slotty stuff in another issue/PR and focus on the changes made here in this thread? Given there is no real solution using fancy interfaces so far, I think this version is the best for now until we find better ways of handling more different snak types in that serializer. Note that the same instanceof check is already done for PropertyValueSnaks so this doesn't really add any new weirdness.

src/Deserializers/SnakDeserializer.php
```diff
@@ -107,6 +107,8 @@ private function newSomeValueSnak( array $serialization ) {
 
 	private function newValueSnak( array $serialization ) {
 		$this->requireAttribute( $serialization, 'datavalue' );
+		// Do not deserialize derived value snaks
+		$this->assertNotAttribute( $serialization, 'derived-values' );
```
JeroenDeDauw: This comment is somewhat misleading, as the line does not prevent deserialization of derived values. You can remove it, and such deserialization would still not happen. What it does is prevent deserializing something that has derived values in it altogether. What is the rationale behind doing that? If I'm not mistaken, we are simply ignoring such things elsewhere.

src/Serializers/SnakSerializer.php
```diff
@@ -81,6 +82,12 @@ private function getSerialized( Snak $snak ) {
 			$serialization['datavalue'] = $this->dataValueSerializer->serialize( $snak->getDataValue() );
 		}
 
+		if ( $snak instanceof DerivedPropertyValueSnak ) {
+			foreach ( $snak->getDerivedDataValues() as $key => $dataValue ) {
+				$serialization['derived-values'][$key] = $this->dataValueSerializer->serialize( $dataValue );
```
JeroenDeDauw: I know there is no error, though should this not be done?

`$serialization['derived-values'] = array();`

src/Deserializers/SnakDeserializer.php
Benestar: I did that exactly because we do it elsewhere: https://github.com/wmde/WikibaseDataModelSerialization/blob/master/src/Deserializers/TermDeserializer.php#L50

addshore:
I think we should continue discussing a solution that fits all cases in the DataModel.
This slotty stuff, or we add more specific objects, we need sitelinks with urls (and maybe more in the future), terms with fallback data, derived snaks etc.

Benestar:
This is still a valid option and should be considered as well.

Benestar:
@brightbyte you might want to have a look at this as well ;-)

thiemowmde:
Sorry if I missed something, this question was probably asked before: What's the point of all this if nobody can read the additional values? What's the point of something that's basically a one-way "dev/null" serialization, not supported and even actively blocked by the deserializers? In my opinion this must be an option that's off by default, so you can use the deserializer both ways, either throwing an exception or not. But if you go that far I have to ask: why is this even in this component and not in a separate validation component that depends on this one?

So this basically boils down to the same issue and same solution as in https://github.com/wmde/WikibaseDataModelSerialization/pull/163#discussion_r38393161 and https://github.com/wmde/WikibaseDataModelSerialization/pull/171#discussion_r38567825: why not have this code in a separate component? I find this so obvious. What do I miss?

Benestar:
After thinking about this again, I think we should support deserializing these values in the deserializers as well, either with the extra values attached or only the plain values. This should be done consequently though for all extra values we have (these are TermFallback and AliasGroupFallback for now).

src/Serializers/SnakSerializer.php
```diff
@@ -81,6 +82,12 @@ private function getSerialized( Snak $snak ) {
...
+			}
```
brightbyte: The proposed structure was a bit different, see https://phabricator.wikimedia.org/T93747

JeroenDeDauw:
From 2015 with no clear present need, so closing. Please reopen if wrong.

PR #163, opened 2015-07-30
Introduce Slotty stuff

addshore:
Just an idea now to discuss

### DIFF
--- a/src/Serializers/SlottySerializer.php
+++ b/src/Serializers/SlottySerializer.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Wikibase\DataModel\Serializers;
+
+use Serializers\DispatchableSerializer;
+use Serializers\Exceptions\SerializationException;
+use Wikibase\DataModel\Slotty;
+
+/**
+ * Package private
+ *
+ * @licence GNU GPL v2+
+ * @author Adam Shorland
+ * @since 1.9.0
+ */
+class SlottySerializer implements DispatchableSerializer {
+
+	/**
+	 * @since 3.0
+	 *
+	 * @param mixed $object
+	 *
+	 * @return boolean
+	 */
+	public function isSerializerFor( $object ) {
+		return $object instanceof Slotty;
+	}
+
+	/**
+	 * @since 1.0
+	 *
+	 * @param Slotty $object
+	 *
+	 * @return array|int|string|bool|float A possibly nested structure consisting of only arrays and scalar values
+	 * @throws SerializationException
+	 */
+	public function serialize( $object ) {
+		return $object->getSlots();
+	}
+}
--- a/src/Slotty.php
+++ b/src/Slotty.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Wikibase\DataModel;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Adam Shorland
+ * @since 1.9.0
+ */
+interface Slotty {
+
+	/**
+	 * @return array
+	 *     Keys: The name of the slot.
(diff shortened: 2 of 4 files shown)


src/Slotty.php
```diff
@@ -0,0 +1,21 @@
...
+ * @since 1.9.0
+ */
+interface Slotty {
+
+	/**
+	 * @return array
+	 *     Keys: The name of the slot.
+	 *     Values: The value of the slot.
+	 *         array[]|int[]|string[]|bool[]|float[]
+	 *         A possibly nested structure consisting of only arrays and scalar values
```
Benestar: I thought those would be DataValue objects?

src/Serializers/SlottySerializer.php
```diff
@@ -0,0 +1,40 @@
...
+use Serializers\DispatchableSerializer;
+use Serializers\Exceptions\SerializationException;
+use Wikibase\DataModel\Slotty;
+
+/**
+ * Package private
+ *
+ * @licence GNU GPL v2+
+ * @author Adam Shorland
+ * @since 1.9.0
```
JeroenDeDauw: wow, 3 different since tags. I guess this supports our mission of diversity!

src/Slotty.php
addshore: Well, I guess they could be, but I see no reason they couldn't just be strings.
If we had them as DataValues and then pushed to use this in the API for urls of sitelinks and datatypes or snaks etc that would require a breaking change in the API or more hacks.

src/Serializers/SlottySerializer.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class SlottySerializer implements DispatchableSerializer {
```
Benestar: I wonder why we need a serializer for this if the object's "slots" are just returned as they exist. If serialization of slots should be that simple, just get rid of that serializer and use `$object->getSlots()` directly.

Benestar:
This way we need to introduce a subclass of `SiteLink` called sth like `SlottySiteLink` that implements the interface and a thing for snaks as well, also for terms. These objects should then become part of this serialization component.

src/Serializers/SlottySerializer.php
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * @since 1.0
+	 *
+	 * @param Slotty $object
+	 *
+	 * @return array|int|string|bool|float A possibly nested structure consisting of only arrays and scalar values
+	 * @throws SerializationException
+	 */
+	public function serialize( $object ) {
+		return $object->getSlots();
```
thiemowmde: I do have two major problems with this proposal:

1. The name is [killing me](http://dict.leo.org/?search=slutty).
2. What's the point of having support for "slots" that can contain nothing but scalar values? What this basically means is: All code working with these "slots" must do it's own (de)serialization to make sure getSlots only returns values that are already in the correct intermediate array format for the JSON and XML serialization. The knowledge about how to (de)serialize a specific "slot" will be spread across different components. If you do this the other way around and have one or many actual "SlotSerializers" that know how to (de)serialize all supported "slots", what's the difference and advantage compared to an actual change to the DataModel and (de)serializers then?

TL;DR: The effect of this proposal is that (de)serializer code will not be in a (de)serializer component any more, but copy-pasted all over the code base where one of the fields must be (de)serialized.

JeroenDeDauw: I have similar concerns already written down on another PR that was doing the same or something very similar. Can't find it now though

JeroenDeDauw:
Assuming this can be closed. Reopen if not

PR #164, opened 2015-08-25
Reduce mentions of Claim(s) without breaking anything

thiemowmde:
This is a non-breaking change that should have zero effect to the public interface. The idea is to reduce the occurrences of the word "claim" in this component to a minimum.

### DIFF
--- a/src/Serializers/ClaimsSerializer.php
+++ b/src/Serializers/ClaimsSerializer.php
@@ -6,8 +6,8 @@ use Serializers\DispatchableSerializer;
 use Serializers\Exceptions\SerializationException;
 use Serializers\Exceptions\UnsupportedObjectException;
 use Serializers\Serializer;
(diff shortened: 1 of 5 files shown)


src/Serializers/ClaimsSerializer.php
```diff
@@ -20,7 +20,7 @@ class ClaimsSerializer implements DispatchableSerializer {
 	/**
 	 * @var Serializer
 	 */
-	private $claimSerializer;
+	private $statementSerializer;
```
JeroenDeDauw: Considering that this is package private, and that it'll be removed once the deprecated `newClaimSerializer` gets removed, you are likely putting in effort that will not help us with anything

thiemowmde: A full text search in the code base always brings these "false" positives up. Increasing code quality even in deprecated code does help me and others. I think this is worth it.

PR #165, opened 2015-08-26
Remove not needed dependency on ByPropertyIdGrouper

JeroenDeDauw:
Needed for https://phabricator.wikimedia.org/T110344

### DIFF
--- a/src/Serializers/StatementListSerializer.php
+++ b/src/Serializers/StatementListSerializer.php
@@ -6,7 +6,6 @@ use Serializers\DispatchableSerializer;
 use Serializers\Exceptions\SerializationException;
 use Serializers\Exceptions\UnsupportedObjectException;
 use Serializers\Serializer;
-use Wikibase\DataModel\Services\ByPropertyIdGrouper;
 use Wikibase\DataModel\Statement\StatementList;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Serializers/StatementListSerializer.php
```diff
@@ -69,12 +68,14 @@ public function serialize( $object ) {
...
+		foreach ( $statementList->toArray() as $statement ) {
+			$idSerialization = $statement->getPropertyId()->getSerialization();
 
-		foreach ( $byPropertyIdGrouper->getPropertyIds() as $propertyId ) {
-			$serialization[$propertyId->getSerialization()] = $this->getSerializedStatements(
-				$byPropertyIdGrouper->getByPropertyId( $propertyId )
-			);
+			if ( !array_key_exists( $idSerialization, $serialization ) ) {
+				$serialization[$idSerialization] = array();
+			}
```
JeroenDeDauw: It works without this if block... not sure if that is good style though, or if it will continue to work in later PHP versions

thiemowmde: Personally I think it's ok to omit this. As long as the thing is initialized as a (1-dimensional) array, PHP will happily add more dimensions when needed. This is tested, so changes in later PHP versions will instantly become visible.

PR #166, opened 2015-08-26
Drop dependency on Wikibase DataModel Services

JeroenDeDauw:


src/DeserializerFactory.php
```diff
@@ -19,7 +19,7 @@
 use Wikibase\DataModel\Deserializers\StatementListDeserializer;
 use Wikibase\DataModel\Deserializers\TermDeserializer;
 use Wikibase\DataModel\Deserializers\TermListDeserializer;
-use Wikibase\DataModel\Services\EntityId\EntityIdParser;
+use Wikibase\DataModel\Entity\EntityIdParser;
```
Benestar: This is a breaking change which results in tests being broken on Wikibase.git.

PR #167, opened 2015-08-27
Revert "Drop dependency on Wikibase DataModel Services"

Benestar:
Reverts wmde/WikibaseDataModelSerialization#166

This was a breaking change and should not get released as a 1.9 but a 2.0. We must do a 1.9.1 release which fixes this issue and then release the changes again as 2.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,12 @@ The recommended way to use this library is via [Composer](http://getcomposer.org
 To add this package as a local, per-project dependency to your project, simply add a
 dependency on `wikibase/data-model-serialization` to your project's `composer.json` file.
 Here is a minimal example of a `composer.json` file that just defines a dependency on
-version 1.5 of this package:
+version 1.0 of this package:
 
 ```json
 {
 	"require": {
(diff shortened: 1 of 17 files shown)


addshore:
Bah, looking at this again yes this is indeed a breaking change
As the namespace has changed things such as the constructor of DeserializerFactory is not expecting a different object..

thiemowmde:
I wonder why we are now so alarmed about a breaking change in https://github.com/wmde/WikibaseDataModelSerialization/compare/1.9.0...master, when https://github.com/wmde/WikibaseDataModelSerialization/compare/1.8.0...1.9.0 did the exact same breaking change, just the other way around. Do we need to release an "unbreaking" 1.8.1 too? And do the same for https://github.com/wmde/WikibaseInternalSerialization/?

addshore:
We probably should, users of this library would shout otherwise.
I know that as I am one of those users, but I was already using 1.9..

JeroenDeDauw:
A difference with the break we avoided now was that the one already made coincided with a major release of WB DM that this component was adapted to.

PR #168, opened 2015-08-27
Revert "Revert "Drop dependency on Wikibase DataModel Services""

Benestar:
This reverts commit 6a78b247868e0020fe3f744d666d702efe771e3f.

Conflicts:
	RELEASE-NOTES.md
	mediawiki.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,8 +28,7 @@
 	},
 	"require": {
 		"php": ">=5.3.0",
-		"wikibase/data-model": "~4.0",
-		"wikibase/data-model-services": "~1.0",
+		"wikibase/data-model": "~4.2",
 		"serialization/serialization": "~3.1",
 		"data-values/serialization": "~1.0"
 	},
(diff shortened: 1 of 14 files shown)


thiemowmde:
Please see https://github.com/wmde/WikibaseDataModelSerialization/pull/167#issuecomment-135340476.

Benestar:
@thiemowmde You mean we should rather consider 1.9.0 as a bugfix of 1.8.0 where we did a breaking release accidentially? I don't think that's a good idea as we already use 1.8.0 in Wikibase.git and this release broke jenkins. Making a mistake once doesn't justify doing the same mistake a second time.

thiemowmde:
Please do not turn my arguments against me. I'm asking if we should fork and release a 1.8.1 to fix 1.8.0. We did exactly that for 1.9.0 now. Both 1.8.0 and 1.9.0 are unusable and need the same bugfix.

Benestar:
Sorry @thiemowmde I think there is a misunderstanding. I didn't want to turn any arguments agains you, how can arguments on software ever be against people? :-) I'm ok with forking 1.8.0 and to a 1.8.1 release but I don't know what semver really encourages in that situation. I just think that a breaking release is definitely needed now, whatever else we release as well.

PR #169, opened 2015-08-27
Remove deprecated methods

### LINKED ISSUE #143: Remove methods deprecated in 1.4
This should be done in the next major release.

Benestar:
Fixes #143 

I've verified that none of these methods are used anywhere in Wikibase.git anymore. They should be removed in 2.0


addshore:
Also checked and doesn't appear to be used in WIkibase or Extensions

PR #170, opened 2015-08-27
Release 2.0.0

Benestar:


RELEASE-NOTES.md
```diff
@@ -1,5 +1,11 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.0.0 (2015-08-27)
+
+* Dropped dependence on Wikibase DataModel Services
```
thiemowmde: Can you please fix both lines to say "dependency"?

Benestar: Sure :-)

JeroenDeDauw: I think it's wrong now and was correct before. You can drop a dependency. Or you can drop dependence on something. Not dropping dependency on something.

PR #171, opened 2015-09-01
Create ExtraValuesAssigner and use it in SiteLinkSerializer

Benestar:
Please comment on this idea, see also #163

### DIFF
--- a/src/ExtraValuesAssigner.php
+++ b/src/ExtraValuesAssigner.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Wikibase\DataModel;
+
+/**
+ * Assigns extra values to the serialization of an object
+ * based on the given extra values providers.
+ *
+ * @since 2.1
+ *
+ * @licence GNU GPL v2+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+class ExtraValuesAssigner {
+
+	/**
+	 * @var ExtraValuesProvider[]
+	 */
+	private $extraValuesProviders;
+
+	/**
+	 * @param ExtraValuesProvider[] $extraValuesProviders
+	 */
+	public function __construct( array $extraValuesProviders = array() ) {
+		$this->extraValuesProviders = $extraValuesProviders;
+	}
+
+	/**
+	 * @param array $serialization
+	 * @param mixed $object
+	 * @return array
+	 */
+	public function addExtraValues( $serialization, $object ) {
+		foreach ( $this->extraValuesProviders as $extraValuesProvider ) {
+			if ( $extraValuesProvider->isProviderFor( $object ) ) {
+				$serialization = array_merge(
+					$extraValuesProvider->getExtraValues( $object ),
+					$serialization
+				);
+			}
+		}
+
+		return $serialization;
+	}
+
+}
--- a/src/ExtraValuesProvider.php
+++ b/src/ExtraValuesProvider.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Wikibase\DataModel;
+
+/**
+ * Provides extra values for the serialization of an object.
+ *
+ * @since 2.1
+ *
+ * @licence GNU GPL v2+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+interface ExtraValuesProvider {
+
+	/**
+	 * @param mixed $object
+	 * @return bool
+	 */
+	public function isProviderFor( $object );
+
+	/**
+	 * @param mixed $object
+	 * @return array
+	 */
+	public function getExtraValues( $object );
+
+}
--- a/src/SerializerFactory.php
+++ b/src/SerializerFactory.php
@@ -57,23 +57,23 @@ class SerializerFactory {
 
 	/**
 	 * @param Serializer $dataValueSerializer serializer for DataValue objects
-	 * @param ExtraValuesAssigner $extraValuesAssigner
 	 * @param int $options set multiple with bitwise or
+	 * @param ExtraValuesAssigner|null $extraValuesAssigner
 	 *
 	 * @throws InvalidArgumentException
 	 */
 	public function __construct(
 		Serializer $dataValueSerializer,
-		ExtraValuesAssigner $extraValuesAssigner,
-		$options = 0
+		$options = 0,
+		ExtraValuesAssigner $extraValuesAssigner = null
 	) {
 		if ( !is_int( $options ) ) {
 			throw new InvalidArgumentException( '$options must be an integer' );
 		}
 
 		$this->dataValueSerializer = $dataValueSerializer;
-		$this->extraValuesAssigner = $extraValuesAssigner;
 		$this->options = $options;
+		$this->extraValuesAssigner = $extraValuesAssigner ?: new ExtraValuesAssigner();
 	}
 
 	/**
--- a/src/SerializerFactory.php
+++ b/src/SerializerFactory.php
@@ -51,17 +51,28 @@ class SerializerFactory {
 	private $dataValueSerializer;
 
 	/**
+	 * @var ExtraValuesAssigner
+	 */
+	private $extraValuesAssigner;
+
+	/**
 	 * @param Serializer $dataValueSerializer serializer for DataValue objects
+	 * @param ExtraValuesAssigner $extraValuesAssigner
 	 * @param int $options set multiple with bitwise or
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function __construct( Serializer $dataValueSerializer, $options = 0 ) {
+	public function __construct(
+		Serializer $dataValueSerializer,
+		ExtraValuesAssigner $extraValuesAssigner,
+		$options = 0
+	) {
 		if ( !is_int( $options ) ) {
 			throw new InvalidArgumentException( '$options must be an integer' );
 		}
 
 		$this->dataValueSerializer = $dataValueSerializer;
+		$this->extraValuesAssigner = $extraValuesAssigner;
 		$this->options = $options;
 	}
 
@@ -121,7 +132,7 @@ class SerializerFactory {
 	 * @return Serializer
 	 */
 	public function newSiteLinkSerializer() {
-		return new SiteLinkSerializer();
+		return new SiteLinkSerializer( $this->extraValuesAssigner );
 	}
 
 	/**
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -12,7 +12,4 @@ if ( !is_readable( __DIR__ . '/../vendor/autoload.php' ) ) {
 }
 
 $loader = require_once( __DIR__ . '/../vendor/autoload.php' );
-$loader->addClassMap( array(
-	'Tests\Wikibase\DataModel\Serializers\SerializerBaseTest' => __DIR__ . '/unit/Serializers/SerializerBaseTest.php',
-	'Tests\Wikibase\DataModel\Deserializers\DeserializerBaseTest' => __DIR__ . '/unit/Deserializers/DeserializerBaseTest.php'
-) );
+$loader->addPsr4( 'Tests\\Wikibase\\DataModel\\', __DIR__ . '/unit/' );
--- a/tests/unit/FakeExtraValuesProvider.php
+++ b/tests/unit/FakeExtraValuesProvider.php
@@ -1,0 +1,31 @@
+<?php
+
+
+namespace Tests\Wikibase\DataModel;
+
+use Wikibase\DataModel\ExtraValuesProvider;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+class FakeExtraValuesProvider implements ExtraValuesProvider {
+
+	/**
+	 * @param mixed $object
+	 * @return bool
+	 */
+	public function isProviderFor( $object ) {
+		return true;
+	}
+
+	/**
+	 * @param mixed $object
+	 * @return array
+	 */
+	public function getExtraValues( $object ) {
+		return array(
+			'foo' => 'bar'
+		);
+	}
+}
--- a/src/Serializers/SiteLinkSerializer.php
+++ b/src/Serializers/SiteLinkSerializer.php
@@ -6,6 +6,7 @@ use Serializers\DispatchableSerializer;
 use Serializers\Exceptions\SerializationException;
 use Serializers\Exceptions\UnsupportedObjectException;
 use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\ExtraValuesAssigner;
 use Wikibase\DataModel\SiteLink;
 
 /**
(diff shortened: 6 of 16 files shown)


src/SerializerFactory.php
```diff
@@ -51,17 +51,28 @@ class SerializerFactory {
...
+	 * @param ExtraValuesAssigner $extraValuesAssigner
 	 * @param int $options set multiple with bitwise or
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function __construct( Serializer $dataValueSerializer, $options = 0 ) {
+	public function __construct(
+		Serializer $dataValueSerializer,
+		ExtraValuesAssigner $extraValuesAssigner,
+		$options = 0
```
Benestar: We should make that an optional parameter and by that avoid the breaking change.

addshore:
Generally I think this looks nicer than both the DerivedPropertyValue stuff and the Slotty stuff.
This would allow is to remove many more bad things from the ResultBuilder in wikibase

Benestar:
It will even allow us to inject the data types of snaks in a nice way because this approach does not depend on inheritance.

src/ExtraValuesAssigner.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Wikibase\DataModel;
```
Benestar: Should be in the Serializers namespace.

JeroenDeDauw: yes

tests/unit/FakeExtraValuesProvider.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+
+namespace Tests\Wikibase\DataModel;
```
JeroenDeDauw: Think I'd be nicer if this is in `fixtures/` rather than `unit/`

JeroenDeDauw:
I'm wondering what real implementations of `ExtraValuesProvider` would look like and where they would be defined. Without knowing that I can't tell if this actually solves any problem.

JeroenDeDauw:
Assuming it will be like

```php
class TypedSnakExtraValuesProvider implements ExtraValuesProvider {
    function getExtraValues( $snak ) {
        return [ 'datatype'  => $this->propertyDataTypeLookup->gimmeh( $snak->getPropertyId() ) ];
    }
}
```

Doing that still violates the contract of serializers, as they are not just serializing the object they are getting. The lookups might not be in the serializer object directly, though will still be dependencies and invoked by it.

Things gained are the ability to use lookups and other things the serializer component does not need to know about and the ability to confugure if and what extra values should be added. However to fully reap those benefits, we need to actually put serialization stuff outside of this component, which a similar concerns as that of @thiemowmde at https://github.com/wmde/WikibaseDataModelSerialization/pull/163#discussion_r38393161

Benestar:
The idea is of course that only the interface is defined here and all the implementations go into a more specific repository (perhaps Wikibase Repo). I believe that his approach is much better than the one using interfaces or extending the value objects because it doesn't need inheritance. We can have multiple ExtraValueProviders for snaks eg. so we can easily assign both a datatype and extended values for PropertyValueSnaks without having to deal with several extending/implementing different interfaces and end up with a bunch of redundant classes.

Using a lookup in the interface isn't really an issue because that's nothing the serialization component knows about. We can also do the lookups before serialization and just let the ExtraValuesProvider get the data from an associative array but I don't see how that would be better. I still think that this solves most if not all of our problems we currently have with serialization.

JeroenDeDauw:
> I believe that his approach is much better than the one using interfaces or extending the value objects because it doesn't need inheritance.

I don't agree with (part of) the reasoning in this line (which is disjoint from the main discussion here): inheritance is not a bad thing. Without it you have no OOP. Code reuse via inheritance is typically bad, though of course does not occur if you just implement an interface.

JeroenDeDauw:
I favour this approach over messing with the data model just for serialization purposes, mainly because a lot less things will be affected by and come to depend on this approach.

> Using a lookup in the interface isn't really an issue because that's nothing the serialization component knows about.

The concern I raised about using lookups in serializers has nothing to do with direct coupling. It is about the behaviour of the serializer, which is supposed to take its input and serializer it, as opposed to adding things coming from somewhere else as well. The proposed approach does not solve this.

A second concern that is not addressed is that of spreading out serialization format knowledge. If you put `ExtraValuesProvider` implementations in repo, then we'll have part of the serialization format here and part of the format in repo. Furthermore, the part that is in repo will not be usable by anything outside of repo, which is not nice. We could create an extra component that depends on serialization and services, or we could just have the `ExtraValuesProvider` implementations in here (and have serialization depend on services again) to avoid the binding to repo.

addshore:
> We could create an extra component that depends on serialization and services, or we could just have the ExtraValuesProvider implementations in here (and have serialization depend on services again) to avoid the binding to repo.

Both of these sound okay to me really, although it would be a shame to introduce the dependency on services here, but if it is needed for more than the 1 thing it was last linked to in here then fine.

src/ExtraValuesProvider.php
```diff
@@ -0,0 +1,27 @@
...
+
+	/**
+	 * @param mixed $object
+	 * @return bool
+	 */
+	public function isProviderFor( $object );
+
+	/**
+	 * @param mixed $object
+	 * @return array
```
thiemowmde: Underspecified. This **is** a serializer and should be documented as such.

src/ExtraValuesProvider.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Provides extra values for the serialization of an object.
+ *
+ * @since 2.1
+ *
+ * @licence GNU GPL v2+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+interface ExtraValuesProvider {
```
thiemowmde: This is not a provider but a serializer.

src/ExtraValuesProvider.php
Benestar: This interface looks surprisingly similar to the DispatchableSerializer interface. Basically the two concepts are very similar. The only difference is that the DispatchingSerializer chooses one serializer and uses only that while this implementation combines and merges all serializations.

src/ExtraValuesProvider.php
Benestar: This is a provider but also a serializer ;-)

src/ExtraValuesAssigner.php
```diff
@@ -0,0 +1,46 @@
...
+	 * @param array $serialization
+	 * @param mixed $object
+	 * @return array
+	 */
+	public function addExtraValues( $serialization, $object ) {
+		foreach ( $this->extraValuesProviders as $extraValuesProvider ) {
+			if ( $extraValuesProvider->isProviderFor( $object ) ) {
+				$serialization = array_merge(
+					$extraValuesProvider->getExtraValues( $object ),
+					$serialization
```
thiemowmde: This is the code line where it becomes obvious that this new concept is nothing but a serializer. So I have to ask what the benefit of a parallel serializer hierarchy outside of the original serializer hierarchy is? Why not use a DataValueSerializer or something similar to a DataValueSerializer?

tests/bootstrap.php
```diff
@@ -12,7 +12,4 @@
 }
 
 $loader = require_once( __DIR__ . '/../vendor/autoload.php' );
-$loader->addClassMap( array(
-	'Tests\Wikibase\DataModel\Serializers\SerializerBaseTest' => __DIR__ . '/unit/Serializers/SerializerBaseTest.php',
-	'Tests\Wikibase\DataModel\Deserializers\DeserializerBaseTest' => __DIR__ . '/unit/Deserializers/DeserializerBaseTest.php'
-) );
+$loader->addPsr4( 'Tests\\Wikibase\\DataModel\\', __DIR__ . '/unit/' );
```
thiemowmde: Please create a separate patch if this is an intended change.

src/ExtraValuesAssigner.php
Benestar: How would you handle things like the url of a sitelink, the datatype of a snak or the page id + revision of an entity using a DataValueSerializer? I think we need a more general concept.

src/ExtraValuesProvider.php
thiemowmde: SerializationProvider then?

Benestar: Could also be ExtraValuesSerializer, tbh I don't care that much how this interface is called. Initially I called it FooBar because I didn't find a better way ^^

src/ExtraValuesAssigner.php
thiemowmde: A concept to inject undefined stuff into a well defined serialization? I think this should be done outside of this components code, after the original serialization is done. It's an array of scalars. We know that. How hard is it to inject something into this array at the right position?

Benestar: So you think the way we currently inject things into the serialization is the best thing we can do? Currently we do exactly that: we first let the serializer do its stuff and then put our things into it. However, people seem to be not very happy with that situation. Otherwise we would not spend so much time on this and have had tickets for this on our sprints.

thiemowmde: I'm talking about an actual, well readable and easy to understand and maintain infrastructure to inject stuff into a serialization. What we currently do is far from that.

Benestar: Honestly I have no idea how we could achieve a readable and easy to understand infrastructure without hooking into the serializers directly like done here. If you have a better idea please share but I don't know any.

thiemowmde: Create a `SerializationInjector` interface with an `inject` method. Create specific implementations for stuff you want to inject. Use them. Done.

Benestar: But that would still require hardcoding the paths of the serialization in some other place outside of this component, right? While I think your idea is definitely worth consideration, it would imo be nicer to not hardcode paths in the serialization. Also this would even more create a "parallel serializer hierarchy outside of the original serializer hierarchy".

In theory when using this serialization component we should only know that the result is an associative array that can be passed to json_encode but nothing more about the content of this array.

thiemowmde: I want a clear cut. The injection code can even go into it's own component, independent from anything but this one. My problem with the approach in this pull request is that:

1. It encodes knowledge about the places **where** stuff can be injected into this component.
2. But **what** can be injected and **how** it must be serialized and deserialized is outside of this component.

I do not think the first issue is the biggest we need to solve. Hard-code some paths. So what? They can not change anyway without becoming a big fat breaking change. This can easily be solved by making the injection component depend on this one.

The problem with the approach in this patch is that you can have breaking changes in the serialization in two different places. If you change something in one of the "extra value" providers it immediately creates a breaking change, but there is no composer dependency in that direction.

JeroenDeDauw:
@Jonas-WMDE I'm curious if you can think of an approach that does not have the remaining issues. If you want I can explain in person what the problem we are trying to solve is in case that is not clear.

thiemowmde:
@JeroenDeDauw, are you referring to something specific @Jonas-WMDE wrote? Or did you meant to ask me? I would like to talk with you about the conclusion I came up with above.

JeroenDeDauw:
@thiemowmde I am asking Jonas for input, as unlike for you and most others in the team, I've not heard his thoughts on this yet. This is completely unrelated to what you wrote above, which to be honest, I'm not quite sure I understand.

src/ExtraValuesAssigner.php
addshore: > The problem with the approach in this patch is that you can have breaking changes in the serialization in two different places. If you change something in one of the "extra value" providers it immediately creates a breaking change, but there is no composer dependency in that direction.

That is also the situation we are currently in... all of this stuff is current done in Repo code. Having a solution like this will make it easier to understand.
Having it in a separate component would make it versionable.

JeroenDeDauw: > Having a solution like this will make it easier to understand.

How?

> It would also increase the efficiency of our serialization code when adding 'all of the things' we currently do in the API.

Can you elaborate?

> Having it in a separate component would make it version-able.

True, though you can just put the current code into its own component as well no?

addshore: > > Having a solution like this will make it easier to understand.

 > How?

Well, do you currently understand what is happening in the ResultBuilder in Repo? :P This would make it 1000 times clearer.

 > > It would also increase the efficiency of our serialization code when adding 'all of the things' we currently do in the API.

 > Can you elaborate?

Currently to add all of the other stuff in that we need through the api for example we do lots of looping through the serialization, as we cant add things to it when it is initially created.
I am guessing that this approach would be more efficient, but I am fairly certain loosing the loops would help.

 > > Having it in a separate component would make it version-able.

 > True, though you can just put the current code into its own component as well no?

That is probably fairly true, although the current stuff binds to mediawiki things, although this could probably be avoided, although it would probably involve something very similar to the approach in this PR, but perhaps still with lots of looping (if we don't like this PR)

brightbyte:
I seem to remember that we had this debate before. I'll try to summarize:

Use cases:
* URLs for sitelinks
* URIs for ID snaks
* Normalized values for Quantity and Time snaks
* (terms with fallback info - not covered by all approaches)

Options:
1. Have the serializers know about the possibility of "extra information" pulled in via service interfaces. The serializer decides how that extra info is to be serialized.
2. Have separate model objects for that allow for the extra info (e.g. a "slotty" PropertyValueSnak, with "slots" for extra values), and have the serializers know about these. 
3. Inject additional information into the serialization array structure after the serializers have run. The injectors need to either be invoked on each level of serialization, or they need to know how to navigate the array structure to find the right point for injection. They need knowledge about the structure generated by the serializers.

I originally went for approach 1 in the old code; if we use interfaces to avoid direct coupling, the main remaining issue with this approach is the conceptual oddity of serializers injecting external knowledge. Approach 2 has the advantage of also covering the use case of terms with fallback info, which isn't covered by the other approaches; but it clutters the data model, and needs extra knowledge in the serializers, typically using instanceof checks.

Approach 3 seems brittle, using the amorphous "serialized"  structure of nested arrays as an interface between different parts of the code base. It also either needs knowledge and cooperation from the serializers (to invoke injectors on every level) or a complex mechanism for finding the correct places to inject specific information. Furthermore, this approach would mean to unserialize parts of the data again, e.g. to construct EntityIds that are needed to look up the information to inject.

I suggest to have a brief meeting to resolve this question.

addshore:
I would love to say it would be great to have this little meeting before I leave tomorrow, but I feel that is unlikely, I will be leaving 30 mins into the story time meeting.

JanZerebecki:
We should document what additional information there is and what it means (i.e. how these relate to the Abstract Data Model).
We should spell out what someone who needs to deserialize and work with these values in PHP should do (i.e. where normally they would use our DataModel PHP binding, but now need to use the additional information).

JeroenDeDauw:
I wonder what happened with the mentioned meeting.

This PR has been inactive since 2015, can it be closed?

JeroenDeDauw:
Actually, just reopen it if not.

PR #174, opened 2015-11-30
Code Experiment: Facet serialization.

brightbyte:
Adds facet (de)serialization infrastructure, and support for the
LanguageFallbackInfo facet for Term.

NOTE: needs unreleased version 5 of wikibase/data-model,
see https://github.com/wmde/WikibaseDataModel/pull/576

NOTE: missing tests for new functionality & classes

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.1 (dev)
+
+* Added support for facets
+
 ## 2.0.0 (2015-08-30)
 
 * Dropped dependency on Wikibase DataModel Services
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
 	},
 	"require": {
 		"php": ">=5.3.0",
-		"wikibase/data-model": "~4.2",
+		"wikimedia/assert": "0.2.2",
+		"wikibase/data-model": "~5.0",
 		"serialization/serialization": "~3.1",
 		"data-values/serialization": "~1.0"
 	},
--- a/src/DeserializerFactory.php
+++ b/src/DeserializerFactory.php
@@ -7,6 +7,7 @@ use Deserializers\DispatchableDeserializer;
 use Deserializers\DispatchingDeserializer;
 use Wikibase\DataModel\Deserializers\AliasGroupListDeserializer;
 use Wikibase\DataModel\Deserializers\EntityIdDeserializer;
+use Wikibase\DataModel\Deserializers\FailingDeserializer;
 use Wikibase\DataModel\Deserializers\ItemDeserializer;
 use Wikibase\DataModel\Deserializers\PropertyDeserializer;
 use Wikibase\DataModel\Deserializers\ReferenceDeserializer;
@@ -19,6 +20,7 @@ use Wikibase\DataModel\Deserializers\StatementListDeserializer;
 use Wikibase\DataModel\Deserializers\TermDeserializer;
 use Wikibase\DataModel\Deserializers\TermListDeserializer;
 use Wikibase\DataModel\Entity\EntityIdParser;
+use Wikibase\DataModel\Serializers\FacetContainerDeserializer;
 
 /**
  * Factory for constructing Deserializer objects that can deserialize WikibaseDataModel objects.
@@ -163,7 +165,17 @@ class DeserializerFactory {
 	 * @return Deserializer
 	 */
 	public function newTermDeserializer() {
-		return new TermDeserializer();
+		//TODO: get facet deserializers from a central registry or hook, so extensions can add their own
+		$facetDeserializers = array(
+			// Depending on context, we may or may not allow language fallback
+			//'LanguageFallbackInfo' => new LanguageFallbackInfoDeserializer(),
+			'LanguageFallbackInfo' => new FailingDeserializer(
+				array( 'source' ),
+				'Refused deserialization of terms with language fallback applied'
+			),
+		);
+
+		return new TermDeserializer( new FacetContainerDeserializer( $facetDeserializers ) );
 	}
 
 	/**
--- a/src/Deserializers/FacetContainerDeserializer.php
+++ b/src/Deserializers/FacetContainerDeserializer.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Wikibase\DataModel\Serializers;
+
+use Deserializers\DispatchableDeserializer;
+use Wikibase\DataModel\Facet\FacetContainer;
+use Wikimedia\Assert\Assert;
+
+/**
+ * Package private
+ *
+ * @since 2.1
+ *
+ * @licence GNU GPL v2+
+ * @author Daniel Kinzler
+ */
+class FacetContainerDeserializer  {
+
+	/**
+	 * @var DispatchableDeserializer[] by facet name
+	 */
+	private $deserializers;
+
+	/**
+	 * @param DispatchableDeserializer[] $deserializers
+	 */
+	public function __construct( array $deserializers ) {
+		Assert::parameterElementType( 'Deserializers\DispatchableDeserializer', $deserializers, '$deserializers' );
+
+		$this->deserializers = $deserializers;
+	}
+
+	/**
+	 * @param FacetContainer $object The object to add facets to.
+	 * @param array $data Serialized data to deserialize
+	 */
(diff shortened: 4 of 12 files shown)


composer.json
```diff
@@ -28,7 +28,8 @@
 	},
 	"require": {
 		"php": ">=5.3.0",
-		"wikibase/data-model": "~4.2",
+		"wikimedia/assert": "0.2.2",
+		"wikibase/data-model": "~5.0",
```
JeroenDeDauw: Use `dev-Facts` as version

src/Deserializers/FacetContainerDeserializer.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Package private
+ *
+ * @since 2.1
+ *
+ * @licence GNU GPL v2+
+ * @author Daniel Kinzler
+ */
+class FacetContainerDeserializer  {
```
JeroenDeDauw: This name confused me, as this actually does not implement the deserializer interface

src/Deserializers/FacetContainerDeserializer.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Wikibase\DataModel\Serializers;
```
JeroenDeDauw: wrong

src/Deserializers/FailingDeserializer.php
```diff
@@ -0,0 +1,59 @@
...
+		Assert::parameterType( 'string', $message, '$message' );
+
+		$this->fields = $fields;
+		$this->message = $message;
+	}
+
+	/**
+	 * @param mixed $serialization
+	 *
+	 * @return Term
```
JeroenDeDauw: ?

src/Deserializers/FailingDeserializer.php
```diff
@@ -0,0 +1,59 @@
...
+	 * @throws DeserializationException
+	 */
+	public function deserialize( $serialization ) {
+		foreach ( $this->fields as $field ) {
+			if ( !array_key_exists( $field, $serialization ) ) {
+				return;
+			}
+		}
+
+		throw new DeserializationException( $this->message );
```
JeroenDeDauw: So failure happens only when all fields are present? Is that intended?

src/Deserializers/LanguageFallbackInfoDeserializer.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @throws DeserializationException
+	 */
+	public function deserialize( $serialization ) {
+		return $this->getDeserialized( $serialization );
+	}
+
+	/**
+	 * @param array $serialization
+	 *
+	 * @return Term
```
JeroenDeDauw: nope

src/Deserializers/LanguageFallbackInfoDeserializer.php
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @licence GNU GPL v2+
+ * @author Daniel Kinzler
+ */
+class LanguageFallbackInfoDeserializer implements DispatchableDeserializer {
+
+	/**
+	 * @param mixed $serialization
+	 *
+	 * @return Term
```
JeroenDeDauw: nope

src/Deserializers/FacetContainerDeserializer.php
brightbyte: Yet it deserializes. What would you like to call it?

src/Deserializers/FacetContainerDeserializer.php
brightbyte: oops

src/Deserializers/FailingDeserializer.php
brightbyte: This class is here to keep the current behavior when deserializing terms. We currently disallow certain constructs on deserialization, to avoid data loss when round-tripping filtered data. If you ask the API for an entity, and tell it to apply language fallback for labels, then we shouldn't allow such fallback terms to be written back via the API. They are not valid input for editing. So we detect them during deserialization.

src/Deserializers/FailingDeserializer.php
brightbyte: Yes. If you want OR logic, just register multiple FailingDeserializers.

src/Deserializers/LanguageFallbackInfoDeserializer.php
brightbyte: what? oh, the return type...

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.1 (dev)
+
+* Added support for facets
+
```
adrianheine: Actually, this only adds support for the `LanguageFallbackInfo` facet and removes support for `TermFallback`.

adrianheine:
This commit is a limited demonstration of the facet approach, because it does not include the difficult part of injecting and configuring. It just converts the inflexible thing we currently have to facets. I think the resulting code is pretty ok, probably better than what we currently have. It does not demonstrate that facets improve flexibility, though.

On the other hand, it already shows an issue I have with this approach. The serialization component is not explicitly configured, not explicitly asked to do what you want it do. Instead, the form the input takes determines the result. That means as a client you have to make sure your data has exactly the form that makes the serialization component perform the right steps. If you don't want language fallbacks in your serialization and you didn't build the datamodel objects yourself, you have to check which facets are there (on every term / datamodel object) and remove those you don't want to see in your serialization. That basically means that datamodel objects stop being an abstract representation a component can take and pass through.

src/Deserializers/FacetContainerDeserializer.php
```diff
@@ -0,0 +1,46 @@
...
+
+		$this->deserializers = $deserializers;
+	}
+
+	/**
+	 * @param FacetContainer $object The object to add facets to.
+	 * @param array $data Serialized data to deserialize
+	 */
+	public function deserializeFacets( FacetContainer $object, array $data ) {
+		foreach ( $this->deserializers as $name => $deserializer ) {
```
thiemowmde: Why is this an array?

src/Deserializers/FailingDeserializer.php
```diff
@@ -0,0 +1,59 @@
...
+	 */
+	private $fields;
+
+	/**
+	 * @var string
+	 */
+	private $message;
+
+	/**
+	 * @param string[] $fields
```
thiemowmde: I find this class extremely confusing. It called a "failing" deserializer. I would expect it to **always** fail. But it does not when one of the fields is present (not absent but present) in the serialization. What's the use of this? The least thing you need to do is to rename the class (no idea to what) and the $fields array. But you can't name it $expectedFields, because it will not check all fields.

The whole thing feels like it should be a private method and not a class.

src/Deserializers/TermDeserializer.php
```diff
@@ -17,6 +18,18 @@
...
 
 	/**
+	 * @var FacetContainerDeserializer
+	 */
+	private $facetDeserializer;
+
+	/**
+	 * @param FacetContainerDeserializer $facetDeserializer
+	 */
+	public function __construct( FacetContainerDeserializer $facetDeserializer ) {
```
thiemowmde: Is there a null implementation, when a user wants to deserialize without this?

src/DeserializerFactory.php
```diff
@@ -19,6 +20,7 @@
 use Wikibase\DataModel\Deserializers\TermDeserializer;
 use Wikibase\DataModel\Deserializers\TermListDeserializer;
 use Wikibase\DataModel\Entity\EntityIdParser;
+use Wikibase\DataModel\Serializers\FacetContainerDeserializer;
```
adrianheine: Namespace

src/Serializers/LanguageFallbackInfoSerializer.php
```diff
@@ -0,0 +1,55 @@
...
+	/**
+	 * @param Term $term
+	 *
+	 * @return array
+	 */
+	private function getSerialized( LanguageFallbackInfo $term ) {
+		$result = array();
+
+		//FIXME: LanguageFallbackInfo probably doesn't need getActualLanguageCode() at all.
+		//$result['language'] = $term->getActualLanguageCode();
```
adrianheine: This is a breaking change

JeroenDeDauw:
Experiment from 2015. Assuming this can be closed. Please reopen if not

PR #179, opened 2015-12-16
More specific type hints in deserializers

thiemowmde:
1. This patch adds some missing `@throws` tags. PHPStorm complains about incomplete PHPDoc sections.
2. I'm adding a few "hard" `array` types in cases where a variable is guaranteed to be an array.
3. I'm making all "soft" `@param … $serialization` type hints on `deserialize` functions as specific as they can be. Technically, the Deserializer interface says "mixed". But the implementations will fail with exceptions if the type does not match. You can say: These requirements are part of the interface of these classes and should be documented as such.
4. Same for `@return …` on `serialize`. These methods are guaranteed to return exactly these types, even if the interfaces says "mixed".

### DIFF
--- a/src/Deserializers/AliasGroupListDeserializer.php
+++ b/src/Deserializers/AliasGroupListDeserializer.php
@@ -21,7 +21,7 @@ class AliasGroupListDeserializer implements Deserializer {
 	/**
 	 * @see Deserializer::deserialize
 	 *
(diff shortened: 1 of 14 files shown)


Benestar:
Needs manual rebase

thiemowmde:
Rebased.

PR #184, opened 2016-01-17
Use extension registration for MediaWiki credits

filbertkm:
also fixes the broken license link on Special:Version

### DIFF
--- a/mediawiki-extension.json
+++ b/mediawiki-extension.json
@@ -1,0 +1,13 @@
+{
+	"name": "Wikibase DataModel Serialization",
+	"author": [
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
If possible, I think it'd be nice to have this file be named `mediawiki.json` rather than `extension.json`, so it's more clear what it is about.

thiemowmde:
I agree with Jeroen. For actual extensions that do auto-loading is critical and also does make sense to name the file "extension.json". In components that aren't extensions this is (as far as I can see) not necessary and also misleading.

addshore:
*agrees with change to COPYING*

PR #185, opened 2016-02-07
Add support for additional entity serializers and deserializers

Benestar:
This patch adds an optional parameter to the factory constructors that
allows to inject additional serializers and deserializers to handle custom
entity types.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.1.0 (dev)
+
+* Added support for additional entity serializers and deserializers
+
 ## 2.0.0 (2015-08-30)
 
 * Dropped dependency on Wikibase DataModel Services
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
This is not needed. You can simply construct a new `DisplatchingDeserializer` in the "new entity serializer" factory function of your application.

Benestar:
That's true. However, we have to make sure that eg. in Wikibase.git all places that instantiate a `Entity(De)Serializer` know about the additional entity types and not only the places where the factory gets created.

JeroenDeDauw:
Are you saying this factory is getting passed around outside of the main wikibase factory? That's dubious... Though even if that is the case, this is a problem to solve there. Doing it here is shoving dependency injection responsibilities of the wikibase.git apps into a library they use.

Benestar:
Apparently this is relevant: see `Wikibase\Repo\Api\ResultBuilder` which deals with a `SerializerFactory` instance that gets injected.

thiemowmde:
I agree with @JeroenDeDauw that this custom factory stuff should not be here, in this component. This should be done on the application level, where the additional entity types live. It should be possible to wrap the call to `newEntitySerializer` in a `DisplatchingDeserializer` and use that.

PR #187, opened 2016-02-14
Add public item/property (de)serializer factory methods

Benestar:
These methods need to be accessible from outside because we want to create
a Dispatchable(De)Serializer in a place where we know all available entity
types and have access to their serializers. That way we can avoid
unneccessarily nesting dispatching serializers and have a clear place in
our code where that dispatching happends.

Bug: [T126932](https://phabricator.wikimedia.org/T126932)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,10 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.1.0 (dev)
+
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
I don't think it's worth doing this just for not having the nested dispatch, which I think makes no real difference for pretty much everything. However, I think it's nice to be able to get a (De)serializer for just an Item or a Property if that is all you need.

PR #188, opened 2016-02-14
Deprecate newEntity(De)Serializer in factories

Benestar:
Dispatching of serializers per entity type should be done when all entity
types are known instead of here where we only know of item and property.

One could argue that one can dispatch a Dispatching(De)Serializer but that
is worse both performance wise and in terms of software architecture.

Bug: [T126932](https://phabricator.wikimedia.org/T126932)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -4,6 +4,8 @@
 
 * Added `newItemSerializer` and `newPropertySerializer` to `SerializerFactory`
 * Added `newItemDeserializer` and `newPropertyDeserializer` to `DeserializerFactory`
+* Deprecated `SerializerFactory::newEntitySerializer`
+* Deprecated `DeserializerFactory::newEntityDeserializer`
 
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
I don't agree with making this change, as I think it's good to provide these methods.

> should be done when all entity types are known

This is about a class of users of this library, not the library itself.

> worse both performance wise and in terms of software architecture.

I think this does not make a real performance difference. If you think it does, please provide numbers. As to architecture, I think this is entirely fine. Wrapping wrappers is a common thing. And indeed, if your client wraps the deserializer, from the clients standpoint, it won't even know a double wrap happened due to the component boundary. Trying to do this kind of optimization over the boundary is not a good idea IMO.

thiemowmde:
Fully agree to @JeroenDeDauw.

thiemowmde:
https://phabricator.wikimedia.org/T126932 is resolved, so can this be closed?

thiemowmde:
See #212 instead.

PR #192, opened 2016-02-16
Run tests also with DM 5.0 stable and 5.x dev

JeroenDeDauw:


.travis.yml
```diff
@@ -10,9 +10,9 @@ matrix:
       php: 5.3
     - env: DM=@dev
       php: 5.4
-    - env: DM=~4.0@dev
+    - env: DM=~5.0@dev
       php: 5.6
-    - env: DM=~4.0
+    - env: DM=~5.0
       php: 7
     - env: DM=~4.0
```
thiemowmde: Was it intentional that you only updated 2 of the 4 mentions of DM?

JeroenDeDauw: Yes. Want to test against all versions we specify compat with. Else we might not notice breaks. That's the whole reason this stuff is set up to specify versions and not just pull in the latest.

PR #193, opened 2016-02-16
Fix broken tests that do nothing or test wrong classes

thiemowmde:
The sheer amount of broken tests I found purely by looking at PHPUnit's code coverage report is incredible.
* `deserializableProvider` is for testing `DispatchableDeserializer::isDeserializerFor`, but most classes under test do not implement this interface. All these test cases are nothing but dead code. I'm changing the `$this->assertTrue( true )` to something that fails to make this obvious.
* A lot of providers missed an `array()`.
* Some tests run with the wrong class.

With this patch, code coverage is at 99.8%. 2 lines are missing. Both are dead code. But I did not wanted to touch this in the same patch.


Benestar:
Thanks for fixing this!

PR #194, opened 2016-03-09
Add compatibility with DataModel 6.0

Benestar:


tests/integration/EntitySerializationRoundtripTest.php
```diff
@@ -35,41 +35,37 @@ public function testEntitySerializationRoundtrips( Entity $entity ) {
 	}
 
 	public function entityProvider() {
-		$entities = array();
+		$items = array();
```
JeroenDeDauw: This is still an `entityProvider`, so `$entities` seems like a better name... just noticed there is actually a property put in this thing at the end of the function :)

Benestar: Oh yes, I missed that!

thiemowmde:
This looks like the fixes to the test can be merged without depending on DM master.

PR #195, opened 2016-03-11
Release 2.2.0

thiemowmde:
Pinging @Benestar.

Note that the EntityDeserializationCompatibilityTest is in a "slow" group and excluded by default. Fixed in #196.

[Bug: T128468](https://phabricator.wikimedia.org/T128468)

### DIFF
--- a/src/DeserializerFactory.php
+++ b/src/DeserializerFactory.php
@@ -54,7 +54,7 @@ class DeserializerFactory {
 	}
 
 	/**
-	 * Returns a Deserializer that can deserialize Entity objects.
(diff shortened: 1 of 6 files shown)


src/DeserializerFactory.php
```diff
@@ -54,7 +54,7 @@ public function __construct(
 	}
 
 	/**
-	 * Returns a Deserializer that can deserialize Entity objects.
+	 * Returns a Deserializer that can deserialize Item and Property objects.
```
Benestar: Thanks, this change is important!

PR #197, opened 2016-03-17
Use real types in return tags

thiemowmde:
See https://gerrit.wikimedia.org/r/277999.

The argument is very, very simple: These methods are **guaranteed** to return exactly these types of deserializers. A method with the name "newTermDeserializer" can never, for example, return a SnakDeserializer or something else. But the interface currently says that, and lies. The deserializers you get can **not** deserialize everything.

### DIFF
--- a/src/DeserializerFactory.php
+++ b/src/DeserializerFactory.php
@@ -70,7 +70,7 @@ class DeserializerFactory {
 	 *
 	 * @since 2.1
 	 *
-	 * @return Deserializer
+	 * @return ItemDeserializer
 	 */
 	public function newItemDeserializer() {
 		return new ItemDeserializer(
@@ -87,7 +87,7 @@ class DeserializerFactory {
 	 *
 	 * @since 2.1
 	 *
-	 * @return Deserializer
+	 * @return PropertyDeserializer
 	 */
 	public function newPropertyDeserializer() {
 		return new PropertyDeserializer(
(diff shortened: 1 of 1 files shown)


thiemowmde:
Pinging @brightbyte. Note how stupid and broken it is that the arguments given in https://gerrit.wikimedia.org/r/277999 and https://github.com/wmde/WikibaseDataModelSerialization/pull/101 block each other, leading to a stale situation where it's impossible to move in any direction.

JeroenDeDauw:
This goes against https://github.com/wmde/WikibaseDataModelSerialization#library-structure

Boundaries force a degree of type safety loss. This is a trade-off. There is not one correct approach.

I'd like to highlight that when working in a codebase without many good boundaries, and when being used to such a situation, it becomes a lot easier to see a single side of this trade-off.

I'm not going to argue the point with Thiemo, as we already covered the useful ground. While my position on this is not absolute and has changed since 2 years ago, I do not think it is a good idea to change the type hints as proposed here.

thiemowmde:
I'm afraid I don't fully understand the wording of your 2nd point. Are you suggesting to **kill** the useless and painful Deserializer interface?

thiemowmde:
Just wondering, since you did not responded for a month:
> I'm not going to argue the point with Thiemo […]

Does this mean you will not respond any more?

PR #198, opened 2016-03-18
[WIP] Introduce type-safe interfaces for De/Serializers

thiemowmde:
This is an alternative for the discussion in #101/#197 and https://gerrit.wikimedia.org/r/277999.

I wanted to do this for a long time now, and it starts making more and more sense.
- Note that this is a only quick demonstration, only touching the aliases de/serializers. I want to do the same to all de/serializers.
- The implementations move to "internal" namespaces (this is split to a separate commit, which makes the 2nd one much easier to review).
- The original file at the original namespace becomes an interface that's as narrow as possible.
- The De/Serializer interfaces are **not** used any more.
- When you compare this proposal to #101/#197 the only advantage is that the **constructors** do not become public. Otherwise #101/#197 would be sufficient.
- Note that the serializers do not have a reason to throw an exception. They have strict type hints against DataModel objects now. There can not be an undeserializable value in a valid DataModel object.


thiemowmde:
Pinging @JeroenDeDauw and @brightbyte.

Benestar:
If we go into this direction, I wonder if it is actually worth it having interfaces and a factory. We could also just expose the public classes and constructors and thereby remove all the overhead that gets created when using the interface-factory approach.

thiemowmde:
The factory still hides all knowledge of how to construct the de/serializers. That's the whole point of a factory, isn't it? Am I missing something?

tobijat:
This is described as an alternative to https://gerrit.wikimedia.org/r/#/c/277999/ which has been merged. So I assume this is then obsolete, @thiemowmde ?

thiemowmde:
Erm, no. #101/#197 are also mentioned and never discussed on a level where I could follow and understand the reasons and arguments given. Simply put: There is a line of documentation stating (de)serializers must be constructed using the factories. I don't have a problem with that. But I have a problem with the factories lying about the objects they return. Arguing that the `SnakSerializer` class, for example, becomes somehow public simply because it's used in a `@return` tag does not make sense to me. See, you can not construct a new SnakSerializer from that. You still have to call the factory. But if that was the (only?) reason the class was marked as "package private", what else is the reason blocking #101/#197?

PR #199, opened 2016-03-18
Make DispatchableDe/Serializer tests as specific as possible

thiemowmde:
This renames base classes and makes them as specific as possible:
* "DeserializerBaseTest" to "DispatchableDeserializerTest".
* "SerializerBaseTest" to "DispatchableSerializerTest".

All tests on non-dispatchable classes can not use these base classes any more.

I'm also switching to `assertSame` to compare the serialized arrays, which makes the order part of the test. Please do not argue with "order is unspecified in JSON". This component does not know anything about JSON. The serializers in this component output PHP arrays. Order is a thing in PHP arrays.

This is meant as a base for #198. It will make working on #198 much easier.


thiemowmde:
Kindly asking, uh, the original author @Tpt for a review? Maybe @JeroenDeDauw?

PR #205, opened 2017-02-03
Micro optimize StatementDeserializer::deserialize

mariushoch:
This gains us about ~5%, per the bug.

https://phabricator.wikimedia.org/T157013

PR #207, opened 2017-02-07
Inline a few trivial private methods

thiemowmde:
This is a direct follow up for #205.

### DIFF
--- a/src/Deserializers/StatementDeserializer.php
+++ b/src/Deserializers/StatementDeserializer.php
@@ -22,11 +22,11 @@ use Wikibase\DataModel\Statement\Statement;
  */
 class StatementDeserializer implements DispatchableDeserializer {
 
-	private $rankIds = array(
+	private static $rankIds = [
 		'deprecated' => Statement::RANK_DEPRECATED,
 		'normal' => Statement::RANK_NORMAL,
 		'preferred' => Statement::RANK_PREFERRED
-	);
+	];
 
 	/**
 	 * @var Deserializer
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
This gains us what? (I think inlining requires justification with numbers)

thiemowmde:
Another 1% gain in runtime, faster and easier to read code (e.g. native `is_string` instead of assertIsString).

src/Deserializers/StatementDeserializer.php
```diff
@@ -61,15 +61,9 @@ public function __construct(
...
-
-	private function hasType( $serialization ) {
-		return is_array( $serialization ) && array_key_exists( 'type', $serialization );
-	}
-
-	private function hasCorrectType( $serialization ) {
-		return in_array( $serialization['type'], array( 'claim', 'statement' ) );
+		return is_array( $serialization )
+			&& array_key_exists( 'type', $serialization )
+			&& ( $serialization['type'] === 'claim' || $serialization['type'] === 'statement' );
```
bekh6ex: IMHO, It was more clear before.

thiemowmde: I disagree. What puzzled me the most was the code that checked if a variable "has type" and "has correct type". What "type"? Does this check if the variable is an array? How can such a type check be "correct"?

If you find `in_array( $serialization['type'], [ 'claim', 'statement' ] )` more readable we can change the last line.

JeroenDeDauw: The method was badly named. `hasTypeKey` would be better.

JeroenDeDauw:
I think it's a bad idea to inline code to get a 1% runtime gain, which simply is not noticeable once you factor in i/o for nearly all usecases. For performance issues I'd also start by looking at what actually annoys the users. The speed of a maintenance script does likely not rate high there. And much bigger gains might be made in other ways (such as not having ones code run on shitty old PHP versions).

manicki:
+1
I am not convinced by the impact of this but I not opposing.
I'd be also fine with renaming that private method instead.

src/Deserializers/StatementDeserializer.php
thiemowmde: I reworked this method into one that is called `isValidClaimType` now.

manicki:
Still looks fine to me after the last change.

PR #208, opened 2017-02-07
Merge duplicate array loops

thiemowmde:
This patch does have two effects:
* It merges duplicate loops that looped the exact same arrays twice.
* This allows to avoid a few duplicate condition checks, e.g. duplicate `is_string`.

This is a direct follow up for #205.

### DIFF
--- a/src/Deserializers/SnakListDeserializer.php
+++ b/src/Deserializers/SnakListDeserializer.php
@@ -37,7 +37,9 @@ class SnakListDeserializer implements Deserializer {
 	 * @return SnakList
 	 */
 	public function deserialize( $serialization ) {
-		$this->assertHasGoodFormat( $serialization );
+		if ( !is_array( $serialization ) ) {
+			throw new DeserializationException( 'The SnakList serialization should be an array' );
+		}
 
 		return $this->getDeserialized( $serialization );
 	}
@@ -52,6 +54,10 @@ class SnakListDeserializer implements Deserializer {
 
 		foreach ( $serialization as $key => $snakArray ) {
 			if ( is_string( $key ) ) {
+				if ( !is_array( $snakArray ) ) {
+					throw new DeserializationException( "The snaks per property \"$key\" should be an array" );
+				}
+
 				foreach ( $snakArray as $snakSerialization ) {
 					/** @var Snak $snak */
 					$snak = $this->snakDeserializer->deserialize( $snakSerialization );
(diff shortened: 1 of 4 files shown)


thiemowmde:
@mariushoch @JeroenDeDauw I suggest to merge at least this (and possibly #207) before tagging the new release (which is already prepared via #209).

thiemowmde:
Benchmarked locally. Performance gain is a bit more than 3%.

JeroenDeDauw:
3% when just calling the deserialization code? What about in any real world use case where the data needs to be read from somewhere? I imagine it is no longer significant then

thiemowmde:
Even if it's only 1% in a real world scenario, 1% equals 20 minutes, according to https://phabricator.wikimedia.org/T157013#2993060. Under these circumstances I don't think it's worth arguing what the value of duplicate code is.

src/Deserializers/SnakListDeserializer.php
```diff
@@ -52,6 +54,10 @@ private function getDeserialized( array $serialization ) {
 
 		foreach ( $serialization as $key => $snakArray ) {
 			if ( is_string( $key ) ) {
+				if ( !is_array( $snakArray ) ) {
+					throw new DeserializationException( 'The snaks per property should be an array' );
```
bekh6ex: Would be nice to mention key here

src/Deserializers/StatementListDeserializer.php
```diff
@@ -52,6 +54,10 @@ private function getDeserialized( array $serialization ) {
 
 		foreach ( $serialization as $key => $statementArray ) {
 			if ( is_string( $key ) ) {
+				if ( !is_array( $statementArray ) ) {
+					throw new DeserializationException( 'The statements per property should be an array' );
```
bekh6ex: Would be nice to mention key in the message

src/Deserializers/SnakListDeserializer.php
bekh6ex: Would be nice to mention key here

thiemowmde: Both cases are old exception messages, and I did not wanted to change them, but I can.

JeroenDeDauw:
Well is it even 1%? If it's 3% just when executing the PHP code I expect it to be way less than 1% when holding i/o into account.

manicki:
+1
I am not convinced by the impact of this but I not opposing.

Issue #200, opened 2016-04-08
Drop PHP 5.3 and 5.4 support


Tpt:
Blocked by wmde/WikibaseDataModel/issues/611

JeroenDeDauw:
I don't see how this would be blocked on DM requirements. If anything, you'd have a block in the other direction. A component can use more recent versions of PHP than its dependencies without any issues.

PR #211, opened 2017-02-08
Bumped PHP requirement to 5.5

JeroenDeDauw:
Fixes https://github.com/wmde/WikibaseDataModelSerialization/issues/200

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ matrix:
   include:
     - env: DM=~6.0
       php: 7.1
-    - env: DM=~5.0
-      php: 5.5
     - env: DM=@dev
+      php: 5.5
+    - env: DM=~5.0
       php: 5.6
     - env: DM=~4.2
       php: 7
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,10 @@ matrix:
   fast_finish: true
   include:
     - env: DM=~6.0
-      php: 5.3
-    - env: DM=~6.0
-      php: 5.4
-    - env: DM=@dev
-      php: 5.5
+      php: 7.1
     - env: DM=~5.0
+      php: 5.5
+    - env: DM=@dev
       php: 5.6
     - env: DM=~4.2
       php: 7
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,11 +1,13 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.4.0 (dev)
+
+* Dropped support for PHP 5.3 and PHP 5.4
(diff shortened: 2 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -1,11 +1,15 @@
...
 ## 2.3.0 (2017-02-07)
 
-* Improved performance of `StatementDeserializer`.
-* Improved type safety throughout the code.
-* Reworked `DeserializerBaseTest` into `DispatchableDeserializerTest`.
-* Reworked `SerializerBaseTest` into `DispatchableSerializerTest`.
+* Improved performance of `StatementDeserializer`
+* Improved type safety throughout the code
+* Renamed `DeserializerBaseTest` to `DispatchableDeserializerTest`
+* Renamed `SerializerBaseTest` to `DispatchableSerializerTest`
```
thiemowmde: This is not correct. It's a very different base class now, not just a rename.

composer.json
```diff
@@ -27,12 +27,13 @@
...
 	},
 	"require": {
-		"php": ">=5.3.0",
+		"php": ">=5.5.0",
 		"wikibase/data-model": "~6.0|~5.0|~4.2",
 		"serialization/serialization": "~3.1",
 		"data-values/serialization": "~1.0"
 	},
 	"require-dev": {
+		"phpunit/phpunit": "~4.8",
```
thiemowmde: Do you mind if I bring this in alphabetic order?

.travis.yml
```diff
@@ -7,12 +7,10 @@ matrix:
...
-      php: 5.3
-    - env: DM=~6.0
-      php: 5.4
-    - env: DM=@dev
-      php: 5.5
+      php: 7.1
     - env: DM=~5.0
+      php: 5.5
+    - env: DM=@dev
       php: 5.6
```
thiemowmde: You switched the two DataModel versions PHP 5.5 and 5.6 use. Not that it matters, just wondering. A special reason to do this?

JeroenDeDauw: no

composer.json
JeroenDeDauw: no

RELEASE-NOTES.md
JeroenDeDauw: Right... "reworked" is so ambiguous tho

Perhaps this should not even be here - if it is an implementation detail the users should not worry about (no breaking change) then why have it here?

manicki:
+1
Looks OK. I am just confused how this PHP/DM version matrix should look like so I'd better tet @thiemowmde have his eye on it.

PR #212, opened 2017-02-09
[DNM] Make factories support custom entity types

thiemowmde:
**Do not merge!** I figured this is the wrong approach. See https://gerrit.wikimedia.org/r/338749 instead.

~~This is more than a convenient method, but solves an actual issue. This allows us to construct a SerializerFactory (and pass it around) that is able to serialize more than the basic entity types.~~

Please review this along with https://github.com/wmde/WikibaseInternalSerialization/pull/110. Both changes must be done together, or not. As @JeroenDeDauw hinted below the architecture proposed in these two patches might be flawed. Could it be that we are just confused by these duplicate (De)SerializerFactories? That other component is called "internal". What does this mean? Could it be that the only thing we need to do is to rename these components and classes to make it more obvious what they represent?

[Bug: T157596](https://phabricator.wikimedia.org/T157596)
[Bug: T157960](https://phabricator.wikimedia.org/T157960)

### DIFF
--- a/src/SerializerFactory.php
+++ b/src/SerializerFactory.php
@@ -3,8 +3,10 @@
 namespace Wikibase\DataModel;
 
 use InvalidArgumentException;
+use Serializers\DispatchableSerializer;
 use Serializers\DispatchingSerializer;
 use Serializers\Serializer;
+use UnexpectedValueException;
 use Wikibase\DataModel\Serializers\AliasGroupListSerializer;
 use Wikibase\DataModel\Serializers\AliasGroupSerializer;
 use Wikibase\DataModel\Serializers\ItemSerializer;
@@ -41,9 +43,9 @@ class SerializerFactory {
 	const OPTION_SERIALIZE_REFERENCE_SNAKS_WITHOUT_HASH = 8;
 
 	/**
-	 * @var int
+	 * @var DispatchableSerializer[]|callable[]
 	 */
-	private $options;
+	private $entitySerializers;
 
 	/**
 	 * @var Serializer
@@ -51,16 +53,31 @@ class SerializerFactory {
 	private $dataValueSerializer;
 
 	/**
+	 * @var int
+	 */
+	private $options;
+
+	/**
+	 * @since 3.0 added the required $entitySerializers argument
+	 *
+	 * @param DispatchableSerializer[]|callable[] $entitySerializers A list of either
+	 *  DispatchableSerializer objects that are able to serialize Entities, or callables that return
+	 *  such objects. The callables must accept the SerializerFactory as first argument.
 	 * @param Serializer $dataValueSerializer serializer for DataValue objects
 	 * @param int $options set multiple with bitwise or
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function __construct( Serializer $dataValueSerializer, $options = 0 ) {
+	public function __construct(
+		array $entitySerializers,
+		Serializer $dataValueSerializer,
+		$options = self::OPTION_DEFAULT
+	) {
 		if ( !is_int( $options ) ) {
 			throw new InvalidArgumentException( '$options must be an integer' );
 		}
 
+		$this->entitySerializers = $entitySerializers;
 		$this->dataValueSerializer = $dataValueSerializer;
 		$this->options = $options;
 	}
@@ -94,15 +111,29 @@ class SerializerFactory {
 	}
 
 	/**
-	 * Returns a Serializer that can serialize Item and Property objects.
-	 *
-	 * @return Serializer
+	 * @throws UnexpectedValueException
+	 * @return DispatchableSerializer A dispatching serializer that can serialize whatever the
+	 *  entity serializers given in the constructor support. There is no default support for Items
+	 *  or Properties.
 	 */
 	public function newEntitySerializer() {
-		return new DispatchingSerializer( array(
-			$this->newItemSerializer(),
-			$this->newPropertySerializer()
-		) );
+		$serializers = [];
+
+		foreach ( $this->entitySerializers as $serializer ) {
+			if ( is_callable( $serializer ) ) {
+				$serializer = call_user_func( $serializer, $this );
+			}
+
+			if ( !( $serializer instanceof DispatchableSerializer ) ) {
+				throw new UnexpectedValueException(
+					'Expected a DispatchableSerializer or a callable returning one'
+				);
+			}
+
+			$serializers[] = $serializer;
+		}
+
+		return new DispatchingSerializer( $serializers );
 	}
 
 	/**
--- a/tests/integration/EntityDeserializationCompatibilityTest.php
+++ b/tests/integration/EntityDeserializationCompatibilityTest.php
@@ -23,6 +23,14 @@ class EntityDeserializationCompatibilityTest extends \PHPUnit_Framework_TestCase
 
 	protected function setUp() {
 		$deserializerFactory = new DeserializerFactory(
+			[
+				function ( DeserializerFactory $factory ) {
+					return $factory->newItemDeserializer();
+				},
+				function ( DeserializerFactory $factory ) {
+					return $factory->newPropertyDeserializer();
+				},
+			],
 			new DataValueDeserializer(
 				array(
 					'string' => 'DataValues\StringValue',
--- a/tests/integration/EntitySerializationRoundtripTest.php
+++ b/tests/integration/EntitySerializationRoundtripTest.php
@@ -23,8 +23,26 @@ class EntitySerializationRoundtripTest extends \PHPUnit_Framework_TestCase {
 	 * @dataProvider entityProvider
 	 */
 	public function testEntitySerializationRoundtrips( EntityDocument $entity ) {
-		$serializerFactory = new SerializerFactory( new DataValueSerializer() );
+		$serializerFactory = new SerializerFactory(
+			[
+				function ( SerializerFactory $factory ) {
+					return $factory->newItemSerializer();
+				},
+				function ( SerializerFactory $factory ) {
+					return $factory->newPropertySerializer();
+				},
+			],
+			new DataValueSerializer()
+		);
 		$deserializerFactory = new DeserializerFactory(
+			[
+				function ( DeserializerFactory $factory ) {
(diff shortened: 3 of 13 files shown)


src/SerializerFactory.php
```diff
@@ -106,6 +107,29 @@ public function newEntitySerializer() {
 	}
 
 	/**
+	 * @since 2.3
+	 *
+	 * @param DispatchableSerializer[]|callable[] $dispatchableSerializers A list of either
+	 *  DispatchableSerializer objects, or callbacks that return such objects.
+	 *
+	 * @return Serializer A dispatching serializer that can serialize whatever the give serializers
```
manicki: give_n_

src/SerializerFactory.php
```diff
@@ -106,6 +107,29 @@ public function newEntitySerializer() {
...
+	 *  support. There is no default support for Items or Properties.
+	 */
+	public function newDispatchingSerializer( array $dispatchableSerializers ) {
+		foreach ( $dispatchableSerializers as &$serializer ) {
+			if ( is_callable( $serializer ) ) {
+				$serializer = call_user_func( $serializer, $this );
+			}
+
+			if ( !( $serializer instanceof DispatchableSerializer ) ) {
+				throw new InvalidArgumentException( '' );
```
manicki: Shouldn't there be some message?

src/SerializerFactory.php
```diff
@@ -94,15 +111,25 @@ private function shouldSerializeReferenceSnaksWithHash() {
 	}
 
 	/**
-	 * Returns a Serializer that can serialize Item and Property objects.
-	 *
-	 * @return Serializer
+	 * @throws IllegalValueException
+	 * @return Serializer A dispatching serializer that can serialize whatever the entity
```
manicki: Shouldn't this then say `@return DispatchingSerializer`? Or is it only for compliance with the interface?

src/SerializerFactory.php
```diff
@@ -41,26 +43,41 @@ class SerializerFactory {
...
 	 * @param Serializer $dataValueSerializer serializer for DataValue objects
 	 * @param int $options set multiple with bitwise or
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function __construct( Serializer $dataValueSerializer, $options = 0 ) {
+	public function __construct(
+		array $entitySerializers,
+		Serializer $dataValueSerializer,
+		$options = 0
```
manicki: As suggested in the other patch, this could also default to self::OPTION_DEFAULT so that users don't have to wonder what is this zero about.

src/SerializerFactory.php
manicki: (nitpick) Also possibly @return could come before @throws, I think this is the order we often have those.

thiemowmde: DispatchingSerializer is not private and we could expose it here. But there is not much value in this. The only difference is that this makes the `addSerializer` method visible. This should not be needed.

I'm always putting `@throws` before `@return`, because it reflects the actual control flow: code can either throw exceptions for different reasons, or return something in the end.

thiemowmde:
I continued working on this and believe this is finished now. The only detail still missing is a specific test scenario for the new constructor parameter.

I also learned that the architecture I build here does have a flaw: you need a (De)SerializerFactory to create an array with an Item and Property(De)Serializer, which is needed in the constructor of the (De)SerializerFactory. In other words: you need two factories.

Possible consequence: drop the support to pass in (De)Serializer objects and only support callables. What do you think?

brightbyte:
@thiemowmde wrote:
> Possible consequence: drop the support to pass in (De)Serializer objects and only support callables. What do you think?

Yes, just support callables.

src/SerializerFactory.php
```diff
@@ -41,26 +43,41 @@ class SerializerFactory {
...
 	 * @param Serializer $dataValueSerializer serializer for DataValue objects
 	 * @param int $options set multiple with bitwise or
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function __construct( Serializer $dataValueSerializer, $options = 0 ) {
+	public function __construct(
+		array $entitySerializers,
+		Serializer $dataValueSerializer,
+		$options = self::OPTION_DEFAULT
```
brightbyte: I'm not quite sure that $options should be in the constructor here. May also be a parameter of newEntitySerializer.

brightbyte:
"Add SerializerFactory::newDispatchingSerializer" is no longer true, as far as I can tell.

src/SerializerFactory.php
manicki: I'm not sure if this is what @brightbyte meant but you could have optional $options parameters in each newXyzSerializer method (or at least in those that make use of options), instead of setting $options in the constructor. I am not sure how these options are actually used (ie. at what point it makes more sense to specify options) so I have no opinion on this.

tests/integration/EntityDeserializationCompatibilityTest.php
```diff
@@ -23,6 +23,14 @@ class EntityDeserializationCompatibilityTest extends \PHPUnit_Framework_TestCase
...
 	protected function setUp() {
 		$deserializerFactory = new DeserializerFactory(
+			[
+				'item' => function ( DeserializerFactory $factory ) {
+					return $factory->newItemDeserializer();
+				},
+				'property' => function ( DeserializerFactory $factory ) {
+					return $factory->newPropertyDeserializer();
+				},
+			],
```
bekh6ex: Does not matter. Can be replaced with `[]`

tests/integration/EntitySerializationRoundtripTest.php
```diff
@@ -23,8 +23,26 @@ class EntitySerializationRoundtripTest extends \PHPUnit_Framework_TestCase {
...
-		$serializerFactory = new SerializerFactory( new DataValueSerializer() );
+		$serializerFactory = new SerializerFactory(
+			[
+				'item' => function ( SerializerFactory $factory ) {
+					return $factory->newItemSerializer();
+				},
+				'property' => function ( SerializerFactory $factory ) {
+					return $factory->newPropertySerializer();
+				},
+			],
```
bekh6ex: Does not matter. Can be replaced with `[]`

tests/integration/EntitySerializationRoundtripTest.php
```diff
@@ -23,8 +23,26 @@ class EntitySerializationRoundtripTest extends \PHPUnit_Framework_TestCase {
...
+		);
 		$deserializerFactory = new DeserializerFactory(
+			[
+				'item' => function ( DeserializerFactory $factory ) {
+					return $factory->newItemDeserializer();
+				},
+				'property' => function ( DeserializerFactory $factory ) {
+					return $factory->newPropertyDeserializer();
+				},
+			],
```
bekh6ex: Does not matter. Can be replaced with `[]`

tests/unit/DeserializerFactoryTest.php
```diff
@@ -15,7 +15,14 @@
...
 
 	private function buildDeserializerFactory() {
-		return new DeserializerFactory( new DataValueDeserializer(), new BasicEntityIdParser() );
+		return new DeserializerFactory( [
+			'item' => function ( DeserializerFactory $factory ) {
+				return $factory->newItemDeserializer();
+			},
+			'property' => function ( DeserializerFactory $factory ) {
+				return $factory->newPropertyDeserializer();
+			},
```
bekh6ex: Only `testNewEntityDeserializer` depends on key values `item` and `property`, other tests don't care.

I guess this setup should be moved to `testNewEntityDeserializer` and this method might use `[]` instead.

tests/unit/SerializerFactoryTest.php
```diff
@@ -28,7 +28,14 @@
...
 
 	private function buildSerializerFactory() {
-		return new SerializerFactory( new DataValueSerializer() );
+		return new SerializerFactory( [
+			'item' => function ( SerializerFactory $factory ) {
+				return $factory->newItemSerializer();
+			},
+			'property' => function ( SerializerFactory $factory ) {
+				return $factory->newPropertySerializer();
+			},
```
bekh6ex: Does not matter. Can be replaced with `[]`

thiemowmde:
> Does not matter. Can be replaced with []

I don't know why you say that. I guess you did not run all tests when you tried this. Please check again and be more specific with your edit request.

bekh6ex:
I checked out the code and run all the tests. They don't fail if it is replaced with []

Best regards,
Aleksey.

-----Original Message-----

JeroenDeDauw:
Can't really review this one... part of a bigger approach which I think is fundamentally flawed. Will bring that up again soonish at a better place.

thiemowmde:
> part of a bigger approach which I think is fundamentally flawed (and counterproductive). […] I'm a bit sadened no one else is seeing this

Do you realize your comment does not give the slightest hint at what you mean? Can you please clarify?

thiemowmde:
> I […] run all the tests. They don't fail if it is replaced with []

You said that already. As I said you must have done something wrong. When I run the tests they fail when there are no entity (de)serializers. Which makes sense because the tests in question are actually testing the entity (de)serialization.

> part of a bigger approach which I think is fundamentally flawed (and counterproductive). […] I'm a bit sadened no one else is seeing this

Do you realize your comment does not give the slightest hint at what you mean? Can you please clarify?

manicki:
> > **Will bring that up again soonish at a better place.** 

> Do you realize your comment does not give the slightest hint at what you mean? Can you please clarify?

I don't think @JeroenDeDauw thinks it does not make sense to explain this here, and I believe it is fine as he thinks the whole approach is wrong, and want to make a more general remark, not just one details of this particular PR.

I am patiently looking forward to what he is going say in that better place :)

It is not impossible that the approach is flawed. We're not going to have a code with this approach released in this very moment, so discussing it and fixing possible improvements seems like a very good idea to me.

bekh6ex:
> I checked out the code and run all the tests. They don't fail if it is replaced with []

Sorry, my mistake. Was running wrong set of tests.

thiemowmde:
Wrong approach, this component is the base DataModel serialization, and not supposed to know about other entity types. See https://gerrit.wikimedia.org/r/338749 instead.

PR #213, opened 2017-02-13
Release 2.3.0

thiemowmde:
I suggest to merge everything in https://github.com/wmde/WikibaseDataModelSerialization/milestone/7 (#206, #207 and #208) and tag an early release 2.3.0.

Then merge #212 and tag a braking 3.0.0 release.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,8 +1,8 @@
 # Wikibase DataModel Serialization release notes
 
-## 2.3.0 (dev)
+## 2.3.0 (2017-02-13)
 
-* Improved performance of `StatementDeserializer`
+* Improved performance of `StatementDeserializer` as well as other deserializers
 * Improved type safety throughout the code
 * Dropped support for PHP 5.3 and PHP 5.4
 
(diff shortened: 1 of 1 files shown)


manicki:
https://github.com/wmde/WikibaseDataModelSerialization/pull/206 is in, I am not sure though about the micro-optimization ones https://github.com/wmde/WikibaseDataModelSerialization/pull/207 and https://github.com/wmde/WikibaseDataModelSerialization/pull/208.
I have no problem with them (I am not sure how much we win with those but they make the code somewhat faster, yes) but it would be nice to have a nod from @JeroenDeDauw if he is fine with them, or if he would like to see hard numbers on how much impact they have on whole Wikibase etc.
I am probably going to merge those if there is nothing happening out there in some more time, though.

JeroenDeDauw:
I'm very skeptical of those changes. However I don't think merging them really makes a difference either way. This is very local and not significant compared to the real issues in the wider Wikibase codebase. What I do think would be harmful is continuing with such inlining and doing so based on wishful thinking.

manicki:
Thanks @JeroenDeDauw, I share scepticism.
I believe starting from version 2.3/3.0 of this component we should try to provide hard data to justify one or other optimization.

Issue #210, opened 2017-02-08
Add suggest section in composer.json


JeroenDeDauw:
With the DataValue libraries used by default by Wikibase Repo

PR #216, opened 2017-03-03
Added suggest section to composer.json w/ DV components

JeroenDeDauw:
Fixes https://github.com/wmde/WikibaseDataModelSerialization/issues/210


JeroenDeDauw:
Fail... that should be descriptions and not version numbers

PR #218, opened 2017-03-15
Release 2.4.0

thiemowmde:


JeroenDeDauw:
This updated requirement kinda sucks. Is it really needed? If the only compat issue here is in the tests of this component, then you can resolve it there, without effecting users of this component that do not care about the hash.

thiemowmde:
I also was not happy with this. I had an other look and believe I found a really nice solution. What do you think?

tests/unit/Serializers/SnakSerializerTest.php
```diff
@@ -19,7 +19,7 @@
 class SnakSerializerTest extends DispatchableSerializerTest {
 
 	protected function buildSerializer() {
-		return new SnakSerializer( new DataValueSerializer() );
+		return new SnakSerializer( new DataValueSerializer(), false );
```
bekh6ex: Could you name this `false`? Otherwise it quiet hard to read

thiemowmde: Done.

PR #219, opened 2017-03-20
Remove unused hashes from JSON dumps

thiemowmde:
These are unused and actually different when using this code with DataModel 7.

### DIFF
--- a/tests/data/items/Q1.json
+++ b/tests/data/items/Q1.json
@@ -1114,7 +1114,6 @@
 				"rank": "normal",
 				"references": [
 					{
-						"hash": "004ec6fbee857649acdbdbad4f97b2c8571df97b",
 						"snaks": {
 							"P143": [
 								{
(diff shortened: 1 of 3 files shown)


thiemowmde:
Rebased, just to run the tests again. @Ladsgroup, do you think we can merge this before doing a release?

thiemowmde:
No, the deserializers do not do anything with these hashes. I don't think doing so (comparing the hash and failing if it does not match) would be a good idea, because the hash algorithm might change for reasons entirely unrelated to anything relevant to the deserialization.

JeroenDeDauw:
Uh, you seem to have misunderstood what I was getting at. Less ambiguous version:

Are these **files/serializations** not used by deserialization tests? If they are, they make sure that old serializations can still be deserialized.

thiemowmde:
The only usage of these files is in EntityDeserializationCompatibilityTest. The intention is indeed to make sure these old serializations can still be deserialized. I was under a wrong impression when I uploaded this patch: since the tests still succeed without these hashes, I assumed they are unused. And they are. But this is not the point. Future deserializers might start doing something with these hashes. The deserialization of old entities must still succeed then.

PR #220, opened 2017-03-23
[DNM] Remove confusing newEntity(De)Serializer methods

thiemowmde:
My original suggestion was to rename this instead of removing it, but this is more or less the same than removing it and reintroducing a new method with a better name when we need it. It looks like we are not gonna need it.

Currently these two methods are still used in a handful of tests. In all these cases it can easily be replaced with either a DispatchingDeserializer, or by splitting the test cases just as I did it in this patch.

[Bug: T160436](https://phabricator.wikimedia.org/T160436)


thiemowmde:
I uploaded a deprecation patch as #222.

PR #221, opened 2017-03-24
Rewrite EntitySerializationRoundtripTest

thiemowmde:
This is a pretty straight refactoring of this test. The only actual change is that this is not testing the `newEntity…` methods any more, but the two separate `newItem…` and `newProperty…`. I believe this is ok for a test that claims to test roundtrips.

There is one aspect that should be tested somewhere: an Item serialization should not be deserialized by the Property deserializer, and the other way around. But this is not the job of this test. It is not even build to test this, and would only catch such an error by random chance.

This is split from #220.


thiemowmde:
Manually rebased. @Ladsgroup, maybe?

PR #226, opened 2017-07-28
Add .gitattributes to exclude not needed files from git exports

manicki:


manicki:
Travis build failures due to broken config. Fixed in https://github.com/wmde/WikibaseDataModelSerialization/pull/227

addshore:
CI fails? O_O :(

PR #228, opened 2017-07-28
Add PHPCS config

manicki:


phpcs.xml
```diff
@@ -0,0 +1,16 @@
...
+	<rule ref="vendor/wikibase/wikibase-codesniffer/Wikibase" />
+
+	<rule ref="Generic.Files.LineLength">
+		<properties>
+			<property name="lineLimit" value="120" />
+		</properties>
+	</rule>
+
+	<rule ref="Generic.Arrays.DisallowLongArraySyntax.Found">
+		<exclude name="Generic.Arrays.DisallowLongArraySyntax.Found" />
```
Ladsgroup: I think we should fix it instead of excluding it, maybe in another PR.

phpcs.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<ruleset name="WikibaseDataModelServices">
+	<rule ref="vendor/wikibase/wikibase-codesniffer/Wikibase" />
+
+	<rule ref="Generic.Files.LineLength">
+		<properties>
+			<property name="lineLimit" value="120" />
```
Ladsgroup: Isn't it easily fixable? if not, nvm but if yes, let's fix it.

manicki: https://github.com/wmde/WikibaseDataModelSerialization/pull/229 :)

Ladsgroup: You are extremely fast.

phpcs.xml
manicki: I didn't count how many issues this found but there were several. I've first raised the limit to 120 and there were only two problems so I chopped those down. I can make it use the default (100 chars AFAIK) but then I am going to be completely ignoring comments like "The diff looks horrible" :)

Ladsgroup:
Travis is not happy about "ci" command, otherwise good to merge.

manicki:
oops, fixed

phpcs.xml
Ladsgroup: Let's do it in another PR

Ladsgroup:
Scrutnizer is crazy and we need to fix that too.

PR #230, opened 2017-08-02
Simplify Composer commands a lot

thiemowmde:


manicki:
I didn't get what you said in the comment/description. Do you want this PR to be merged now or not?

For the actual PR itself: Why do you think it is better to have single composer script, and e.g. not be able to run phpcs separately?
I normally work so I run tests while writing code (to verify stuff works), and only run code style checks before submitting change. With this I'd possibly need to run all of it, on both "stages". Probably not a big deal given the size of the repo but just saying.
So I can see a possible down side of this change. What would be a win then?

JeroenDeDauw:
I also am not sure what the up side of this change is. I've come to expect the ci command to be there and am using this format in all my new repo's. Not seen any reason to remove some of the commands or heard anyone raise one so far.

thiemowmde:
* This patch should be merged, but I suggest to wait before tagging the actual `2.5.0` release.
* Currently there are 4 commands. Each executes an unexpected subset. `cs` and `phpcs` even do the exact same. I regret introducing these. They are uncommon, not really predictable and barely useful because of this. No other MediaWiki code repository uses any of these. Not even `ci`. Only very few Wikibase components do, and even these don't use the commands consistently. The only consistent command is `test`, which always (on all MediaWiki extensions and components) executes all tests, including code sniffers. This makes the fact that `composer test` does not execute the code sniffers what is unexpected here.
* I'm fine with keeping `ci` for now. Feel free to amend this patch if you think this helps.

manicki:
> No other MediaWiki code repository uses any of these. 

Not sure whether what MediaWiki code repository do are of high relevance but I note 
https://phabricator.wikimedia.org/diffusion/EWBA/browse/master/composer.json;9ac6d06fb100722b9755a01cdf821fb03a580066$84 for instance

>  They are uncommon

cs and ci scripts are defined in many repositories owned by WMDE as @JeroenDeDauw said. Is it not common enough?

>  not really predictable and 

When running "composer cs" I expect to have all automated code style checks run.
When running "composer ci" I expect to have all automated checks run that will be run as a part of the CI build.

Are those interpretation incorrect? What exactly are you finding not predictable?

> barely useful because of this.

I said in my previous comment why I like having "cs" separated from "test". To add on this, I like "composer cs" over "composer phpcs" as it is shorter. And in some repos cs does more/other than phpcs, I prefer to not have to remember what code style tools are used in repo X and which in repo Y.

lucaswerkmeister:
I certainly agree that for consistency with other MediaWiki stuff, `composer test` should include `phpcs`. Having a separate command to just run the unit tests is definitely valuable (PHPCS is slooow), but it can’t be `composer test` – and since the command is apparently just `phpunit` anyways, I’m not sure if it needs a definition in `composer.json` too.

JeroenDeDauw:
MediaWiki is an application and a framework. This is a library. Why would it need to be consistent with MediaWiki? That is like saying the Fundraising Donation component should be consistent with Symfony because the applications that use it are building on Symfony. This makes absolutely no sense to me.

thiemowmde:
I believe it would be more productive if you provide your own arguments, instead of just dismissing other peoples statements.

JeroenDeDauw:
Irony much?

thiemowmde:
That's not even a sentence. What do you want to say?

I believe it does make sense to give developers tools they can rely on. If the group of developers that owns a set of codebases decides they want `composer test` to always run PHPCS along with PHPUnit, there is value in this decision. It might be that such a decision was not yet made for the particular group in question here. This doesn't mean it's ok to make fun of the idea.

manicki:
The change suggested here is not about, or not only about "composer test always run PHPCS along with PHPUnit", but also about removing composer ci and composer cs commands.
The reasoning for all this is making composer test command in a way consistent with some other MediaWiki-related code bases, am I getting the point right?
While it is good to motivate changes you want to make, it should be also kept in mind that others might question that reasoning, especially if you're calling for the decision made by the group. The consistency argument could be played multiple ways, and if the consistency is the only reason for change, I shrug on this one, if you asked me to contribute to the decision.

What I find not cool is that despite the points were brought about not removing "composer ci" command, you still suggest to remove it and ask others to restore if they like. Those people already said they prefer to have it.

JeroenDeDauw:
Been here a year and has multiple objections

PR #233, opened 2017-08-24
Merge options to use snak hashes into one

lucaswerkmeister:
Suppressing snak hashes was initially added as a single option in 656398540f, but then split up into three options in 679774d120, since the previous lib serializer included qualifier hashes but not main snak or reference hashes, and the goal was to migrate from that serializer to this one with absolutely no change in the output. Now that this migration is done, we can merge the three options into one again, since it’s hardly useful to have snak hashes for some snaks but not others: either include all hashes, or omit all of them.

This is implemented as an extra constant, which is the bitwise union of the three existing ones, for backwards compatibility. With the next major release, the three old options should be removed and the single option made functional instead (with a single method shouldSerializeSnaksWithHash()).

---

> With the next major release, the three old options should be removed

Is there a convenient place in this repository to note this down so we don’t forget it? Should I create an issue with milestone 3.0.0?

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,13 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.5.0 (dev)
+
+* Deprecated `SerializerFactory` options
+  `OPTION_SERIALIZE_MAIN_SNAKS_WITHOUT_HASH`,
+  `OPTION_SERIALIZE_QUALIFIER_SNAKS_WITHOUT_HASH` and
+  `OPTION_SERIALIZE_REFERENCE_SNAKS_WITHOUT_HASH`,
+  and added `OPTION_SERIALIZE_SNAKS_WITHOUT_HASH` instead
+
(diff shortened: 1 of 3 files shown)


lucaswerkmeister:
Note that this makes the PHP 5.5 Travis build fail with a syntax error. It was already failing before, but just with test failures. Is it acceptable to use PHP 5.6 syntax or should I inline the const (14)?

manicki:
Added version number to @deprecated notes in https://github.com/wmde/WikibaseDataModelSerialization/pull/233/commits/b21922ce60aa74a3aca9f7f51907e388c119bc10, I think it makes it clear when this code should be gone.

Looks good, makes sense and I am willing to merge.
Shouldn't we already add shouldSerializeSnaksWithHash in the same release?
We probably also could deprecate shouldSerializeMainSnaksWithHasht and the other two then . What do you think @lucaswerkmeister ?

manicki:
meh, those are private. silly me. never mind then.

PR #237, opened 2017-09-15
Release 2.6.0

thiemowmde:
Note that #222 should be in this release!

I used the opportunity to lower the line length a little bit, to a limit where I only had to touch two lines in a single file. I ended touching a few more lines because I decided to rename some variables (I found $snakDeserializerMock and $snaksDeserializerMock especially confusing – the only difference was an "s" in the middle of the variable name).

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel Serialization release notes
 
+## 2.6.0 (2017-09-15)
+
+* Added compatibility with DataValues Number 0.9
+
 ## 2.5.0 (2017-08-30)
 
 * Removed MediaWiki integration files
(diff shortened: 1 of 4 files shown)


RELEASE-NOTES.md
```diff
@@ -1,14 +1,18 @@
...
 
 * Removed MediaWiki integration files
 * Added compatibility with DataValues Geo 2.x
 * Deprecated `SerializerFactory` options
   `OPTION_SERIALIZE_MAIN_SNAKS_WITHOUT_HASH`,
   `OPTION_SERIALIZE_QUALIFIER_SNAKS_WITHOUT_HASH` and
-  `OPTION_SERIALIZE_REFERENCE_SNAKS_WITHOUT_HASH`,
-  and added `OPTION_SERIALIZE_SNAKS_WITHOUT_HASH` instead
+  `OPTION_SERIALIZE_REFERENCE_SNAKS_WITHOUT_HASH`.
+* Added `SerializerFactory::OPTION_SERIALIZE_SNAKS_WITHOUT_HASH`.
```
manicki: Just leaving a note that this change does not really seem related/necessary to me but whatever.

PR #239, opened 2017-09-19
Document public getters to return Dispatchable(De)serializers

thiemowmde:
To know this is critical for external callers as well as internally. Look how the newEntityDeserializer method uses the return values from newItemDeserializer and newPropertyDeserializer: in a DispatchingDeserializer, which requires the values to be of type DispatchableDeserializer. Same is true for usages outside of this component. So this information is already public, just not documented properly.

### DIFF
--- a/src/SerializerFactory.php
+++ b/src/SerializerFactory.php
@@ -131,7 +131,7 @@ class SerializerFactory {
 	 *
 	 * @since 2.1
 	 *
-	 * @return Serializer
+	 * @return DispatchableSerializer
 	 */
 	public function newItemSerializer() {
 		return new ItemSerializer(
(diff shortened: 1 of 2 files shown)


src/SerializerFactory.php
```diff
@@ -186,7 +186,7 @@ public function newStatementListSerializer() {
 	 *
 	 * @since 1.4
 	 *
-	 * @return Serializer
+	 * @return DispatchableSerializer
 	 */
 	public function newStatementSerializer() {
```
thiemowmde: I touched this because the corresponding deserializer is already documented as "dispatchable".

This is for a historical reason: certain API modules have been designed in a way they can accept both statement as well as claim serializations. Therefor statement and claim serializers and deserializers must be dispatchable. And they are, it's just the documentation that is misleading. I think we should fix this, even if we are not using this as a feature at the moment.

PR #241, opened 2017-10-23
Fix SnakDeserializer causing "Undefined index" errors

thiemowmde:
* The fact that an UnDeserializableValue was constructed just *assuming* the "type" and "value" array elements are there was a bug. This code path is executed when the DataValueDeserializer fails, which includes the situation when one or both array elements are just missing.
* I'm also removing the awkward EntityIdDeserializer injection. This code path can not accept anything but PropertyIds. There was even a check for this. Why do we need to call an overly complex generic "EntityId" deserializer, if we know already what entity type we expect? Note this is not a breaking change because all this is declared "package-private".
* Also remove the unused "composer cs" command.

[Bug: T178651](https://phabricator.wikimedia.org/T178651)

### DIFF
--- a/src/Deserializers/SnakDeserializer.php
+++ b/src/Deserializers/SnakDeserializer.php
@@ -2,6 +2,8 @@
 
 namespace Wikibase\DataModel\Deserializers;
 
+use DataValues\DataValue;
+use DataValues\Deserializers\DataValueDeserializer;
 use DataValues\UnDeserializableValue;
 use Deserializers\Deserializer;
 use Deserializers\DispatchableDeserializer;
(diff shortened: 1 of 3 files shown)


src/Deserializers/SnakDeserializer.php
```diff
@@ -114,12 +108,24 @@ private function newValueSnak( array $serialization ) {
...
+			$value = isset( $serialization[DataValueDeserializer::VALUE_KEY] )
+				? $serialization[DataValueDeserializer::VALUE_KEY]
+				: null;
+			$type = isset( $serialization[DataValueDeserializer::TYPE_KEY] )
+				? $serialization[DataValueDeserializer::TYPE_KEY]
+				: null;
 			$error = isset( $serialization['error'] ) ? $serialization['error'] : $ex->getMessage();
-			return new UnDeserializableValue( $serialization['value'], $serialization['type'], $error );
+
+			return new UnDeserializableValue( $value, $type, $error );
```
mariushoch: Even though the documentation for `UnDeserializableValue` is horrible (phpdoc not at all in line with the actual code), this seems valid.

mariushoch:
+1 looks good to me!

Documentation of `UnDeserializableValue::__construct` is awful, but that has nothing to do with this change.

PR #242, opened 2017-10-23
Update used Wikibase CodeSniffer rule set to version 0.3.0

thiemowmde:


composer.json
```diff
@@ -60,11 +60,8 @@
 			"phpunit"
 		],
 		"ci": [
-			"@cs",
+			"phpcs -p -s",
```
lucaswerkmeister: Is there any special motivation for inlining this? I’m only familiar with WBQC’s `composer.json` scripts, where `test` includes `phpcs`, so since the scripts here are different from that before and after this change, I’m not sure if it’s a good change :)

thiemowmde: The motivation is simplicity and consistency. See #230 for a longer discussion about this. I can remove this change from this patch here if it is in the way.

lucaswerkmeister: Thanks for the link… but since #230 didn’t come to any conclusion as far as I can see, I wouldn’t feel comfortable merging the same change here without giving Jeroen and Leszek an opportunity to respond. I think it would be better to keep that out of this PR for now, and perhaps try to finish up the discussion in #230.

thiemowmde: Done.

lucaswerkmeister: Thank you!

PR #244, opened 2018-02-22
Start using PHPUnit 6 forwards compatibility aliases

thiemowmde:
[Bug: T188011](https://phabricator.wikimedia.org/T188011)


thiemowmde:
Learning from this: Unfortunately the `…|^6` idea can not be used. Travis pulls in PHPUnit 6, and then complains because it can not find the `setExpectedException()` method any more. It's called `expectException()` now. `getMock()` also doesn't exist any more.

PR #250, opened 2018-12-13
Apply ID prefix mapping in SnakDeserializer.

brightbyte:
See https://phabricator.wikimedia.org/T211927

Since SnakDeserializer is documented to be package private, this is not a breaking change. A bug fix release is presumably warranted.

### DIFF
--- a/src/Deserializers/SnakDeserializer.php
+++ b/src/Deserializers/SnakDeserializer.php
@@ -12,7 +12,9 @@ use Deserializers\Exceptions\InvalidAttributeException;
 use Deserializers\Exceptions\MissingAttributeException;
 use Deserializers\Exceptions\MissingTypeException;
 use Deserializers\Exceptions\UnsupportedTypeException;
-use InvalidArgumentException;
+use Wikibase\DataModel\Entity\EntityIdParser;
+use Wikibase\DataModel\Entity\EntityIdParsingException;
+use Wikibase\DataModel\Entity\Property;
 use Wikibase\DataModel\Entity\PropertyId;
 use Wikibase\DataModel\Snak\PropertyNoValueSnak;
 use Wikibase\DataModel\Snak\PropertySomeValueSnak;
@@ -32,8 +34,17 @@ class SnakDeserializer implements DispatchableDeserializer {
 	 */
 	private $dataValueDeserializer;
 
-	public function __construct( Deserializer $dataValueDeserializer ) {
+	/**
+	 * @var EntityIdParser
+	 */
+	private $propertyIdParser;
+
+	public function __construct(
+		EntityIdParser $propertyIdParser,
+		Deserializer $dataValueDeserializer
+	) {
 		$this->dataValueDeserializer = $dataValueDeserializer;
+		$this->propertyIdParser = $propertyIdParser;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


tests/unit/Deserializers/SnakDeserializerTest.php
```diff
@@ -113,6 +115,14 @@ public function deserializationProvider() {
 					'hash' => 'not a valid hash'
 				]
 			],
+			[
+				new PropertyNoValueSnak( new PropertyId( 'foo:P42' ) ),
```
thiemowmde: This line here makes the dataProvider fail. The code you want to test is not reached.

brightbyte: Works for DataModel version 6, but not before. Meh.

thiemowmde: Ah, yea. We could drop this compatibility requirement from travis.yml. Would be fine for me.

brightbyte: If you drop it from travis.yml, also drop it from composer.json.
I removed the test for now, it's not really necessary.

src/Deserializers/SnakDeserializer.php
```diff
@@ -137,8 +148,18 @@ private function deserializeDataValue( $serialization ) {
 	 */
 	private function deserializePropertyId( $serialization ) {
 		try {
-			return new PropertyId( $serialization );
-		} catch ( InvalidArgumentException $ex ) {
+			$id = $this->propertyIdParser->parse( $serialization );
+
+			if ( $id->getEntityType() !== Property::ENTITY_TYPE ) {
```
mariushoch: This is a very very "hot" function… can we save a function call here by switching this out for "instanceof PropertyId"?

brightbyte: I don't know if instanceof is actually faster. Want to benchmark?

mariushoch:
```
$ php maintenance/eval.php --wiki wikidatawiki
> $propertyId = new Wikibase\DataModel\Entity\PropertyId( 'P123' );

> $propertyId->getEntityType();

>  $t0 = microtime(1); for ( $i = 0; $i < 100000; $i++ ) { if ( $propertyId->getEntityType() !== Wikibase\DataModel\Entity\Property::ENTITY_TYPE ) { echo 'fail'; } }; echo microtime(1) - $t0;
0.061592102050781
>  $t0 = microtime(1); for ( $i = 0; $i < 100000; $i++ ) { if ( !$propertyId instanceof Wikibase\DataModel\Entity\PropertyId ) { echo 'fail'; } }; echo microtime(1) - $t0;
0.017740964889526
>  $t0 = microtime(1); for ( $i = 0; $i < 100000; $i++ ) { if ( $propertyId->getEntityType() !== Wikibase\DataModel\Entity\Property::ENTITY_TYPE ) { echo 'fail'; } }; echo microtime(1) - $t0;
0.056143999099731
>  $t0 = microtime(1); for ( $i = 0; $i < 100000; $i++ ) { if ( !$propertyId instanceof Wikibase\DataModel\Entity\PropertyId ) { echo 'fail'; } }; echo microtime(1) - $t0;
0.018042087554932
```

So yes, I think `instanceof` is actually quite a bit faster :)

PR #263, opened 2023-01-08
composer.json: Update IRC Support to libera.chat

reedy:


manicki:
thanks for the suggestion. This library has moved to be maintained in Wikibase "monorepo", and this github repo is a secondary copy. The IRC pointer is however incorrect, so I'll replay your change on Gerrit. Cheers

manicki:
see https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/877094

